PR #192, opened 2023-12-27
NR-199912: Background Instrumentation

cdillard-NewRelic:
// NOTE: BackgroundReporting is only enabled for iOS 13+.

Accomplishes Acceptance Criteria of tickets:

https://new-relic.atlassian.net/browse/NR-238932
https://new-relic.atlassian.net/browse/NR-238931

Prerequisites in your iOS app project to use Background instrumentation

Add the key `Permitted background task scheduler identifiers` with an array including an item containing your apps bundle Id.

Enable `Background Fetch` and `Background Processing` in the “Background Modes” section of the Xcode project capabilities.

Run test harvest in background with:

```
e -l objc -- (void)[[BGTaskScheduler sharedScheduler] _simulateLaunchForTaskWithIdentifier:@"com.newrelic.NRTestApp.bitcode"]
```
replacing `com.newrelic.NRTestApp.bitcode` with your apps bundleId.

Or Await  the periodically occurring harvest in the background.

The exact frequency at which your tasks will run is determined by the system based on factors like device usage, batt life, and workloads. Therefore while we can request periodic execution , the system ultimately is the one that decides when to execute tasks. The background task will be run once per 24hr period.

In order to add BackgroundTasks to app you must add this to your AppDelegate. 
```
    // Background fetch handling.
    func application(_ application: UIApplication, performFetchWithCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void) {

        NewRelic.logVerbose("performFetchWithCompletionHandler called")
        completionHandler(.newData)
    }
```

-------------

Adds changes to NewRelicAgentInternal such that BackgroundTasks framework is interacted with and background task can be scheduled for harvest 

Comments out following files as they are no longer used in our agent project.

Agent/HandledException/NRMAHexBackgroundUploader.h
Agent/HandledException/NRMASessionIdentifierManager.h
Tests/Unit-Tests/NewRelicAgentTests/Handled-Exception-Tests/TestSessionIdentifierManger.m


Agent/General/NewRelicAgentInternal.m
```diff
@@ -149,6 +149,17 @@ - (id) initWithCollectorAddress:(NSString*)collectorHost
...
     self = [super init];
     if (self) {
+
+        if ([NRMAFlags shouldEnableBackgroundReporting]) {
+            if (@available(iOS 13.0, *)) {
+                [[BGTaskScheduler sharedScheduler] registerForTaskWithIdentifier:[[NSBundle mainBundle] bundleIdentifier] usingQueue:nil launchHandler:^(__kindof BGTask * _Nonnull task) {
+                    [self handleAppRefreshTask: task];
+                }];
+            } else {
+                // Fallback on earlier versions
```
smalsam-newr, line 159: Is there an actual fallback for pre-iOS 13 systems? Or do we just not support this feature on those?

Agent/HandledException/NRMAHexBackgroundUploader.mm
```diff
@@ -3,46 +3,46 @@
...
-    [self.session invalidateAndCancel];
-    [self.session release];
-    [self.hexHost release];
-    [self.sessionIdManager release];
-
-    [super dealloc];
-}
-
-@end
+//#import "NRMAHexBackgroundUploader.h"
```
smalsam-newr, line 6: If the contents of this file are going away, it would be better to simply delete it.

Agent/Harvester/NRMAHarvester.mm
```diff
@@ -352,7 +353,19 @@ - (void) connected
 #ifndef  DISABLE_NRMA_EXCEPTION_WRAPPER
     @try {
 #endif
-        [NRMATaskQueue queue:[[NRMAMetric alloc] initWithName:kNRSupportabilityPrefix@"/Collector/Harvest"
+
+        // TODO: Modify this Support Metric to use NewRelicAgentInternal applicationState.
```
smalsam-newr, line 357: It looks like this has been done, so the TODO is no longer relevant.

Agent/General/NewRelicAgentInternal.m
cdillard-NewRelic: BackgroundReporting is only enabled for iOS 13+. Will note this.

Agent/HandledException/NRMAHexBackgroundUploader.mm
cdillard-NewRelic: Deleted.

Agent/General/NewRelicAgentInternal.m
smalsam-newr: In that case, I would just get rid of all the else clauses. There's no need for them.

cdillard-NewRelic: Done

mbruin-NR (approved):

mbruin-NR (approved):

PR #211, opened 2024-03-01
NR-234826 - Missing DT Headers

smalsam-newr:
In several instances, Distributed Tracing headers were not added to requests before they were sent off. This was due to not calling the proper method which would, simultaneously generate the DT headers, and add them to the request. Later, the Distributed Tracing Payload would be associated with the request, allowing for round trip tracing. In addition, DT headers were not being added to Swift Async requests before they went off, due to the delegate method setState not giving us a proper updated state in the newState parameter.

### DIFF
--- a/Agent/Instrumentation/NSURLSession/NRMAURLSessionTaskOverride.m
+++ b/Agent/Instrumentation/NSURLSession/NRMAURLSessionTaskOverride.m
@@ -126,7 +126,7 @@ void NRMAOverride__urlSessionTask_SetState(NSURLSessionTask* task, SEL _cmd, NSU
                 
                 NSURL *url = [currentRequest URL];
                 if (url != nil &&
-                    newState != NSURLSessionTaskStateRunning && task.state == NSURLSessionTaskStateRunning) {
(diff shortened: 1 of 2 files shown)


cdillard-NewRelic:
Should this point at the `develop` branch?

Agent/Instrumentation/NSURLSession/NRMAURLSessionTaskOverride.m
```diff
@@ -126,7 +126,7 @@ void NRMAOverride__urlSessionTask_SetState(NSURLSessionTask* task, SEL _cmd, NSU
                 
                 NSURL *url = [currentRequest URL];
                 if (url != nil &&
-                    newState != NSURLSessionTaskStateRunning && task.state == NSURLSessionTaskStateRunning) {
```
cdillard-NewRelic, line 129: See slack convo, is this change really needed? The intent is to only call contents within if statement once per request.

cdillard-NewRelic (approved):

mbruin-NR (approved):

PR #213, opened 2024-03-07
Added the offline attribute to the old event system

mbruin-NR:
The offline attribute was only added to the new event system. This pr adds the attribute to the old event system. An event should have the offline attribute when it is created while the device is in an offline state.


cdillard-NewRelic (dismissed):
👍 Excellent work and test additions.

smalsam-newr (approved):

PR #214, opened 2024-03-08
NR-239832 - Fix incorrect NSURLSession cast

smalsam-newr:
A crash was found when the allHeaders selector was sent to an instance of NSURLSession that was not also an NSHTTPURLSession, and this cast was not checked.


mbruin-NR (approved):

cdillard-NewRelic (approved):
👍

PR #221, opened 2024-03-21
NR-251740 - DT headers to dataTaskWithRequest call

smalsam-newr:
When adding support for the New Event System, headers were improperly added just to the mutable request, which first was not even used to create the NSURLSessionTask. The headers were added to the mutableRequest after the task was created, meaning that it was not part of the task.


mbruin-NR (approved):

cdillard-NewRelic (commented):
👍 It would be great to add a unit test that replicates the issue RN was seeing. This way we could be sure to never regress DT this way again. I'll try to create a ticket and look into writing and adding this test soon.

PR #228, opened 2024-04-11
Feature/NR-189606: Retrieve Static HTTP Headers

smalsam-newr:
For use in hybrid agents, added an additional static method to retrieve additional tracked headers added to requests.


cdillard-NewRelic (approved):
👍 LGTM

PR #232, opened 2024-04-25
Background Inst: Harvest using gcd

cdillard-NewRelic:
Mike Bruin created a new addition to our sample app. This adds a Picture in Picture video playing functionality to the NRTestApp. Simply tap video view to demonstrate Picture and Picture video functionality and how it interacts with the New Relic IOS agent.

I identified that our current Agent/Harvester/NRMAHarvestTimer.m code is using an NStimer and specifically this code: 

```
[[NSRunLoop mainRunLoop] addTimer:self.timer forMode:NSDefaultRunLoopMode];

```

The issues is that when the app is running in a backgrounded state it stops calling things on the NSRunLoop mainRunLoop. 

This PR makes a change to the NRMAHarvestTimer.m  such that it does not use an NSTimer or the mainRunLoop anymore. It now uses a GCD Dispatch timer, which properly runs when the app is backgrounded.


Agent/Harvester/NRMAHarvestTimer.m
mbruin-NR, line 1: I think this is a good change, but it still needs some of the functionality of the pervious timer. Mainly the ability to stop or cancel it.

PR #238, opened 2024-04-30
NR-259309 added a simple test app for watchOS agent

mbruin-NR:
This pr adds a bare bones test app for the watchOS agent. The test app cannot be built through for the simulator, a physical device is needed.


cdillard-NewRelic (approved):
👍

smalsam-newr (approved):

PR #241, opened 2024-05-09
Added Test Target to watchOS

smalsam-newr:
Adds a test target for the WatchOS agent. Due to limitations of WatchOS, some tests were omitted, and some tests were modified to better suit WatchOS.


cdillard-NewRelic (approved):
👍

PR #243, opened 2024-05-14
Added notifications to NewRelicAgentInternal for watchOS

mbruin-NR:
I was able to figure out how to get watchOS to listen to the notifications for WKApplicationWillEnterForegroundNotification and WKApplicationDidEnterBackgroundNotification.


cdillard-NewRelic (approved):
👍 Looks good, if possible/easy we should add a test but all good either way.

smalsam-newr (approved):

PR #244, opened 2024-05-16
NR-267244 changed the way watchOS determines offline

mbruin-NR:
This pr changes the way that watchOS checks for offline based on how Android does it.


cdillard-NewRelic (approved):
👍 I was able to fix my watch debug connection and test this with physical watchOS device.

PR #256, opened 2024-06-04
NR-270228: New setUserId session handling.

cdillard-NewRelic:
https://new-relic.atlassian.net/browse/NR-270228 

This PR adds new session handling to the setUserId routine.
This accomplishes the following AC:
```
     1. When setUserID(value: string|null) is called:
        a. If userID was previously null and new value is non-null:
            i. continue the current session and set the new userID
        b. If userID was previously not-null and new value is different (including null):
            i. end the current session and perform harvest
            ii. start a new session with the new userID
```

There is challenges testing this change. There will be tests added at a later date.

### DIFF
--- a/Agent/Public/NewRelic.m
+++ b/Agent/Public/NewRelic.m
@@ -602,44 +602,34 @@
 }
 
 + (BOOL) setUserId:(NSString* _Nullable)userId {
-
-    /*
-     
-     1. When setUserID(value: string|null) is called:
-        a. If userID was previously null and new value is non-null:
-            i. continue the current session and set the new userID
-        b. If userID was previously not-null and new value is different (including null):
-            i. end the current session and perform harvest
-            ii. start a new session with the new userID
-     */
     NSString *previousUserId = [[NewRelicAgentInternal sharedInstance] getUserId];
     BOOL newSession = false;
     // If the client passes a new userId that is non NULL.
     if (userId != NULL) {
-        // If userId has not previously been set.
-        if (previousUserId == NULL) {
+
+        // A new userId has been set where the previously set one (during this app session (since app launch))was not NULL.
+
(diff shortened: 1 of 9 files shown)


Agent/Public/NewRelic.m
```diff
@@ -600,10 +601,63 @@ + (BOOL) incrementAttribute:(NSString*)name
...
+    // If the client passes a new NULL userId.
+    else {
+        if (previousUserId == NULL) {
+            // Do nothing if passed user id is null and saved userId is null.
+        }
+        else {
+            // end session and harvest.
+            newSession = true;
+        }
+    }
```
mbruin-NR, line 637: It looks like you wrote the if statements to be deliberate to what was required in the spec but I don't like the empty conditions. Doesn't hurt anything but looks weird.

cdillard-NewRelic: Fixed.

Agent/Public/NewRelic.m
```diff
@@ -600,10 +601,53 @@ + (BOOL) incrementAttribute:(NSString*)name
...
+
+        // A new userId has been set where the previously set one (during this app session (since app launch))was not NULL.
+
+        if (previousUserId != NULL) {
+            // continue session and set the new userID
+            newSession = true;
+        }
+        // If userId has not previously been set this session
+        else {
+            newSession = true;
```
mbruin-NR, line 618: I think this else is not needed.

cdillard-NewRelic: Thanks, I pushed a commit that rectifies this issue by always starting a new Session if the passed userId is non Null.

smalsam-newr (approved):

smalsam-newr (approved):

PR #276, opened 2024-07-15
NR-160576 change to make sure the minimum time is respected.

mbruin-NR:
This PR changes the maxQueueTime detection to account for the 1 min harvest cycle. This change allows events to be harvested within the maxBufferTime instead of the minute after when the harvest cycle ticks.

- Scenario 1: No events in the queue. 
      Events will not be harvested
- Scenario 2: maxBufferTimeSeconds = 60 seconds. 
       As long as an event exists it will be harvested every harvest cycle.(1 min)
- Scenario 3: maxBufferTimeSeconds = 600 seconds. 
       Events will be sent when the oldest event is between 541-600 seconds old at the harvest cycle tick.


smalsam-newr (approved):

cdillard-NewRelic (dismissed):
Looking good. any test update possible?

cdillard-NewRelic (approved):
LGTM and good test updates

PR #282, opened 2024-08-05
Nr 282926 try to make a watchOS offline detection system

mbruin-NR:
This pr tries to make a watchOS offline detection system. It uses a NSURLSessionDataTask to determine if we are offline at the time an event is created without blocking the thread and without spamming the NSURLSessionDataTask to much.


Agent/Analytics/NRMAAnalytics.mm
```diff
@@ -268,7 +268,24 @@ - (BOOL) addInteractionEvent:(NSString*)name
...
+
+#else
+        __block auto event = _analyticsController->addInteractionEvent([name UTF8String], duration_secs, [self checkBackgroundStatus]);
+        
+        [self checkOfflineStatus:^(BOOL isOffline) {
+            if(isOffline){
+                event->addAttribute(kNRMA_Attrib_offline.UTF8String, @YES.boolValue);
+            }
+        }];
+        return _analyticsController->addEvent(event);
```
smalsam-newr, line 287: I think you want to add the event inside your completion handler. Otherwise I'm not sure the offline attribute will consistently be added.

smalsam-newr (approved):

Agent/Analytics/NRMAAnalytics.mm
mbruin-NR: That is ideal, issue with that is that the function returns a bool indicating whether the event was added successfully. Do you know of a way we could maintain that? In the code we don't use that return value it's only used for unit testing.

smalsam-newr: If the agent itself doesn't use it, then we might not needed it. But if you want to continue with it, have your completion handler take a bool indicating that.

mbruin-NR: The return value is used and is important for custom events. I'm still not sure what the code would look like to have the return in the completion handler.

smalsam-newr (approved):

mbruin-NR:
@cdillard-NewRelic I think there would be a way to do that but I don't know how valuable the offline attribute would be when it has such low accuracy. If a customer isn't sure that it means what it says then it might as well not have it.

cdillard-NewRelic:
Ok this PR can be closed then.

mbruin-NR:
This change is not low impact enough to be added to our agent.

PR #283, opened 2024-08-06
NR-289825:  Log Reporting: Console log level should be separate from Remote logging level,  NR-300193: Log Reporting: Use customCollector address

cdillard-NewRelic:
https://new-relic.atlassian.net/browse/NR-289825 

[iOS] [Phase 2] Log Reporting: Console log level should be separate from Remote logging level.

https://new-relic.atlassian.net/browse/NR-300193

[iOS] [phase 2] Log Reporting: Use customCollector address

### DIFF
--- a/Agent/Measurements/NRMAMeasurementType.m
+++ b/Agent/Measurements/NRMAMeasurementType.m
@@ -5,35 +5,3 @@
 //  Created by Bryce Buchanan on 9/26/13.
 //  Copyright © 2023 New Relic. All rights reserved.
 //
-
-#import "NRMAMeasurementType.h"
-
-NSString* NSStringFromNRMAMeasurementType(NRMAMeasurementType type)
-{
-    switch (type) {
-        case NRMAMT_Activity:
-            return @"Activity";
-            break;
-        case NRMAMT_HTTPError:
-            return @"HTTPError";
-            break;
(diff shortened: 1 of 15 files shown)


mbruin-NR:
Is there a way to add unit tests that prove that the log levels are respected differently?

cdillard-NewRelic:
> Is there a way to add unit tests that prove that the log levels are respected differently?

I've added test named testRemoteLogLevels which prove the remote log level respected differently.

mbruin-NR (approved):

Agent/Measurements/NRMAMeasurementType.m
```diff
@@ -5,35 +5,3 @@
 //  Created by Bryce Buchanan on 9/26/13.
 //  Copyright © 2023 New Relic. All rights reserved.
 //
-
```
smalsam-newr, line 8: What was the reason for removing this?

smalsam-newr (approved):

Agent/Measurements/NRMAMeasurementType.m
cdillard-NewRelic: I noticed it wasn't being used but it showed up in a crash stack i was looking at. I figured its best to remove the dead dead code if its a safe removal like this.

PR #287, opened 2024-08-20
NR-303754 attempting to address crash accessing NRMAPayload properties in event creation.

mbruin-NR:
In this pr I moved the retrieving of the payload to before dispatch_async and tried to be more safe in places.


smalsam-newr (approved):

cdillard-NewRelic (approved):
👍

PR #297, opened 2024-08-22
NR-305349 - NRMAAnalytics lazy initialization

smalsam-newr:
Check if NRMAAnalytics is already initialized before blindly re-initializing it again.

This still needs to be tested for background instrumentation.

### DIFF
--- a/Agent/General/NewRelicAgentInternal.m
+++ b/Agent/General/NewRelicAgentInternal.m
@@ -445,7 +445,9 @@ static NSString* kNRMAAnalyticsInitializationLock = @"AnalyticsInitializationLoc
 - (void) initializeAnalytics {
     @synchronized(kNRMAAnalyticsInitializationLock) {
         // [NRMAAnalytics clearDuplicationStores];
-        self.analyticsController = [[NRMAAnalytics alloc] initWithSessionStartTimeMS:(long long)([self.appSessionStartDate timeIntervalSince1970] * 1000)];
+        if(!self.analyticsController) {
+            self.analyticsController = [[NRMAAnalytics alloc] initWithSessionStartTimeMS:(long long)([self.appSessionStartDate timeIntervalSince1970] * 1000)];
+        }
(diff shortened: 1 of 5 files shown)


mbruin-NR (dismissed):

cdillard-NewRelic (dismissed):
👍 tested with BG reporting and other than a double harvest on bg/fg that was preexisting it looks good. Approved.

mbruin-NR (commented):
There is a bug when background mode is enabled. Because the analyticsController never get's destroyed in background mode with this change the analyticsController will never get recreated. Because of this the _sessionStartTime never changes on a new session.
The 'appSessionStartDate' get's updated in 'sessionStartInitialization' and then as part of 'initializeAnalytics' that value is passed into the newly initialized 'analyticsController'. Make sure the _sessionStartTime in the analyticsController reflects a new session.

Agent/General/NewRelicAgentInternal.m
```diff
@@ -445,7 +445,15 @@ - (void) memoryUsageDidUpdate:(NSNotification*)notification {
 - (void) initializeAnalytics {
     @synchronized(kNRMAAnalyticsInitializationLock) {
         // [NRMAAnalytics clearDuplicationStores];
-        self.analyticsController = [[NRMAAnalytics alloc] initWithSessionStartTimeMS:(long long)([self.appSessionStartDate timeIntervalSince1970] * 1000)];
+        if(!self.analyticsController) {
+            self.analyticsController = [[NRMAAnalytics alloc] initWithSessionStartTimeMS:(long long)([self.appSessionStartDate timeIntervalSince1970] * 1000)];
+        } else if(didFireEnterForeground && didFireEnterBackground) {
+            // We are coming back to the foreground after having a background stint
```
mbruin-NR, line 451: There are some comments that can be removed here.

mbruin-NR (approved):

PR #317, opened 2024-10-07
NR-308480 fixed the double mobileSession

mbruin-NR:
Changed the #if so that NSProcessInfo is used for watchOS and not for iOS/tvOS


cdillard-NewRelic (approved):
I've tested this branch on watchOS and IOS and the /data requests appear correct.
This is an excellent catch! Thank you for your hard work! 👍

PR #320, opened 2024-10-15
Nr 318119 auto collect stdout & stderr logs

mbruin-NR:
https://new-relic.atlassian.net/browse/NR-318119

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,7 +16,7 @@ jobs:
     name : TestIOS 
     # runs-on: will be set to macos-latest when running on actual GHA. 
     # *** runs-on: ubuntu-latest is used when running via act on mac os. ***
-    runs-on: macos-13 
+    runs-on: macos-14
     steps:
     - uses: actions/checkout@master
       with: 
@@ -96,7 +96,7 @@ jobs:
     name: DeployS3
     # runs-on: will be set to macos-latest when running on actual GHA. 
     # *** runs-on: ubuntu-latest is used when running via act on mac os. ***
-    runs-on: macos-13 
+    runs-on: macos-14
     needs: [testIOS, testTVOS, testWatchOS]
     if: github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/main'
     outputs:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,7 +24,7 @@ jobs:
     
     - uses: maxim-lobanov/setup-xcode@v1
       with:
-        xcode-version: '14.3'
+        xcode-version: '14.3.1'
 
     - name: Install lcov
       run: |
@@ -52,7 +52,7 @@ jobs:
     
     - uses: maxim-lobanov/setup-xcode@v1
       with:
-        xcode-version: '14.3'
+        xcode-version: '14.3.1'
 
     - name: Install lcov
       run: brew install lcov
@@ -78,7 +78,7 @@ jobs:
       
       - uses: maxim-lobanov/setup-xcode@v1
         with:
-          xcode-version: '14.3'
+          xcode-version: '14.3.1'
 
       - name: Install lcov
         run: brew install lcov
@@ -108,7 +108,7 @@ jobs:
     
     - uses: maxim-lobanov/setup-xcode@v1
       with:
-        xcode-version: '14.3'
+        xcode-version: '14.3.1'
 
     - name: Install the Apple certificate and provisioning profile
       env:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,7 +24,7 @@ jobs:
     
     - uses: maxim-lobanov/setup-xcode@v1
       with:
-        xcode-version: '14.3.1'
+        xcode-version: '15.4'
 
     - name: Install lcov
       run: |
@@ -52,7 +52,7 @@ jobs:
     
     - uses: maxim-lobanov/setup-xcode@v1
       with:
-        xcode-version: '14.3.1'
+        xcode-version: '15.4'
 
     - name: Install lcov
       run: brew install lcov
@@ -78,7 +78,7 @@ jobs:
       
       - uses: maxim-lobanov/setup-xcode@v1
         with:
-          xcode-version: '14.3.1'
+          xcode-version: '15.4'
 
       - name: Install lcov
         run: brew install lcov
@@ -108,7 +108,7 @@ jobs:
     
     - uses: maxim-lobanov/setup-xcode@v1
       with:
-        xcode-version: '14.3.1'
+        xcode-version: '15.4'
 
     - name: Install the Apple certificate and provisioning profile
       env:
--- a/Agent/FeatureFlags/NRMAFlags.m
+++ b/Agent/FeatureFlags/NRMAFlags.m
@@ -47,8 +47,7 @@ static NSString* __deviceIdentifierReplacement = NULL;
                               NRFeatureFlag_NetworkRequestEvents | 
                               NRFeatureFlag_RequestErrorEvents |
                               NRFeatureFlag_DistributedTracing |
-                              NRFeatureFlag_AppStartMetrics |
-                              NRFeatureFlag_AutoCollectLogs;
+                              NRFeatureFlag_AppStartMetrics;
                   });
     return __flags;
 }
@@ -178,10 +177,6 @@ static NSString* __deviceIdentifierReplacement = NULL;
     return ([NRMAFlags featureFlags] & NRFeatureFlag_BackgroundReporting) != 0;
 }
 
-+ (BOOL) shouldEnableAutoCollectLogs {
-    return ([NRMAFlags featureFlags] & NRFeatureFlag_AutoCollectLogs) != 0;
-}
-
 + (NSArray<NSString*>*) namesForFlags:(NRMAFeatureFlags)flags {
     NSMutableArray *retArray = [NSMutableArray array];
     if ((flags & NRFeatureFlag_InteractionTracing) == NRFeatureFlag_InteractionTracing) {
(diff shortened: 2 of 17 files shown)


Agent/Utilities/NRLogger.m
```diff
@@ -346,7 +357,6 @@ - (NSData*) jsonDictionary:(NSDictionary*)message {
 
 
     NSMutableDictionary *providedAttributes = [message mutableCopy];
-    [providedAttributes removeObjectsForKeys:@[NRLogMessageLevelKey,NRLogMessageFileKey,NRLogMessageLineNumberKey,NRLogMessageMethodKey,NRLogMessageTimestampKey,NRLogMessageMessageKey]];
```
cdillard-NewRelic, line 349: The intention of this line is to remove reserved attributes from user provided attributes when passed to NewRelic.logAttributes.

cdillard-NewRelic: Do you still want to remove this line? If so can you explain why?

Test Harness/NRTestApp/NRTestApp/AppDelegate.swift
```diff
@@ -33,7 +33,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                                  NRMAFeatureFlags.NRFeatureFlag_OfflineStorage])
         // Note: Disabled by default. Enable or disable (default) flag to enable background reporting.
         // NewRelic.enableFeatures([NRMAFeatureFlags.NRFeatureFlag_BackgroundReporting])
-
+        
+        // Note: Disabled by default. Enable or disable (default) flag to enable auto collect of stdout.
+        //NewRelic.enableFeatures([NRMAFeatureFlags.NRFeatureFlag_RedirectStdOutStdErr])
```
cdillard-NewRelic, line 38: Sounds like John  would like AutoCollectedLogs to be on by default. But we should double check with him and Android.

Agent/Utilities/NRLogger.m
mbruin-NR: This line doesn't interfere with auto collected logs like I originally thought. Adding it back it.

Agent/FeatureFlags/NRMAFlags.m
```diff
@@ -47,7 +47,8 @@ + (NRMAFeatureFlags) featureFlags
                               NRFeatureFlag_NetworkRequestEvents | 
                               NRFeatureFlag_RequestErrorEvents |
                               NRFeatureFlag_DistributedTracing |
-                              NRFeatureFlag_AppStartMetrics;
+                              NRFeatureFlag_AppStartMetrics |
+                              NRFeatureFlag_AutoCollectLogs;
```
cthomas-newrelic, line 51: Careful with new feature flags: we _try_ to keep FF parity between the agents.

Agent/Utilities/NRAutoLogCollector.h
```diff
@@ -0,0 +1,19 @@
+//
+//  NRAutoLogCollector.h
+//  Agent
+//
+//  Created by Mike Bruin on 10/9/24.
```
cthomas-newrelic, line 5: Generally, we try to avoid putting our names into commits. It's on the commit already. If it's a team practice, I understand.

Agent/Utilities/NRAutoLogCollector.m
```diff
@@ -0,0 +1,168 @@
...
+
+@implementation NRAutoLogCollector
+
++ (BOOL) redirectStandardOutputAndError {
+    if (hasRedirectedStdOut){
+        return true;
+    }
+    // Create pipes for stdout and stderr
+    if (pipe(stdoutPipe) == -1 || pipe(stderrPipe) == -1) {
+        return false;
```
cthomas-newrelic, line 30: Should close the valid pipe if the other returns -1.

Agent/Utilities/NRAutoLogCollector.m
```diff
@@ -0,0 +1,168 @@
...
+        close(stdoutPipe[1]);
+        close(stderrPipe[0]);
+        close(stderrPipe[1]);
+        close(saved_stdout);
+        close(saved_stderr);
+        return false;
+    }
+    close(stdoutPipe[1]); // Close the original write end of the stdout pipe
+    close(stderrPipe[1]); // Close the original write end of the stderr pipe
+
```
cthomas-newrelic, line 56: You should investigate disabling buffering of the pipes

cthomas-newrelic (commented):
Nice job.

Agent/Utilities/NRAutoLogCollector.h
mbruin-NR: It is team practice, mainly because Xcode automatically adds it when you create a new file.

Agent/FeatureFlags/NRMAFlags.m
mbruin-NR: I think it's very important to have this feature flag for iOS even if it's not on Android. This is because it breaks logging to Xcode console. If we want to keep parity I would greatly appriciate Android adding the feature flag and ability to disable auto capturing logs.

Agent/Utilities/NRAutoLogCollector.m
cthomas-newrelic: Yes, I believe that's correct.

mbruin-NR: I tried adding that but it seems to mess with the accuracy of the logs. Giving more logs out of order and having partial logs.

Agent/FeatureFlags/NRMAFlags.m
smalsam-newr: Instead of a feature flag, could we detect when the debugger is attached? I imagine we wouldn't want to collect logs in that instance

cdillard-NewRelic: I think we still need the feature flag, and I believe this auto capture should be disabled by default.  It should also not collect logs when the debugger is attached.

cdillard-NewRelic: The reason being that we like to gate all new features with a feature flag in case there is issues.

Agent/Utilities/NRLogger.m
```diff
@@ -299,19 +316,37 @@ - (NSData*) jsonDictionary:(NSDictionary*)message {
...
-                                         NRLogMessageEntityGuidKey: entityGuid,                                                                                                  // 9
-                                         NRLogMessageInstrumentationProviderKey: NRLogMessageMobileValue,                                                                        // 10
-                                         NRLogMessageInstrumentationNameKey: nativePlatform,                                                                                               // 11
-                                         NRLogMessageInstrumentationVersionKey: [NRMAAgentConfiguration connectionInformation].deviceInformation.agentVersion,                   // 12
-                                         NRLogMessageInstrumentationCollectorKey: nativePlatform};                                                                                         // 13
+    NSMutableDictionary *requiredAttributes = [NSMutableDictionary dictionary];
```
cdillard-NewRelic, line 319: Why was this changed? I am working on some changes in this area. Im concerned that we might miss required attributes with the way this was changed.

Agent/Utilities/NRLogger.m
```diff
@@ -299,19 +316,37 @@ - (NSData*) jsonDictionary:(NSDictionary*)message {
...
+
+    value = [[message objectForKey:NRLogMessageMessageKey]stringByReplacingOccurrencesOfString:@"\"" withString:@"\\\""];
+    if (value) [requiredAttributes setObject:value forKey:NRLogMessageMessageKey]; // 6
+
+    if (NRSessionId) [requiredAttributes setObject:NRSessionId forKey:NRLogMessageSessionIdKey]; // 7
+    if (nrAppId) [requiredAttributes setObject:nrAppId forKey:NRLogMessageAppIdKey]; // 8
+    if (entityGuid) [requiredAttributes setObject:entityGuid forKey:NRLogMessageEntityGuidKey]; // 9
+
+    [requiredAttributes setObject:NRLogMessageMobileValue forKey:NRLogMessageInstrumentationProviderKey]; // 10
+    [requiredAttributes setObject:nativePlatform forKey:NRLogMessageInstrumentationNameKey]; // 11
```
cdillard-NewRelic, line 344:
```suggestion
    [requiredAttributes setObject:name forKey:NRLogMessageInstrumentationNameKey]; // 11
```

Agent/Utilities/NRLogger.m
cdillard-NewRelic: If you'd like to keep the changes, which is could be okay if they aren't required by log API... then lets make the following change:

```
                                         // "instrumentation.name"
                                         NRLogMessageInstrumentationNameKey: name, // 11                                                         
                                         // "collector.name
                                         NRLogMessageInstrumentationCollectorKey: nativePlatform};  // 13   
```

mbruin-NR: I made this change so that the agent wouldn't crash if the value was null. There is still probably a change we could make so that it again won't add a log if it doesn't have the required attributes.

cdillard-NewRelic:
Do we think there is anyway to allow the console and auto collected logs to work at the same time? If not its fine but we should note it in the docs. Thanks!

mbruin-NR:
> Do we think there is anyway to allow the console and auto collected logs to work at the same time? If not its fine but we should note it in the docs. Thanks!

I can't think of a way to put it on the console too, maybe we could try to redirect back to console on every log message but that's not how it's setup. It's going to be part of the docs.

.github/workflows/main.yml
```diff
@@ -16,15 +16,15 @@ jobs:
     name : TestIOS 
     # runs-on: will be set to macos-latest when running on actual GHA. 
     # *** runs-on: ubuntu-latest is used when running via act on mac os. ***
-    runs-on: macos-13
```
smalsam-newr, line 19: Are these updates on the runner because of this feature?

smalsam-newr (approved):

.github/workflows/main.yml
mbruin-NR: Github started failing watchOS and tvOS with the Xcode version it was set to not being present. So I updated it but I know there have been other pr's where that wasn't an issue. Maybe it was fixed on the GitHub side, I can remove it if we want to verify.

cdillard-NewRelic (approved):

Agent/FeatureFlags/NRMAFlags.m
cdillard-NewRelic, line 51: `NRFeatureFlag_AutoCollectLogs` must be disabled by default. As my resolved comment indicated. Thanks!

PR #321, opened 2024-10-17
Pass nativePlatform as instrumentation insights name, collector.name …

cdillard-NewRelic:
…is hybrid paltform.


mbruin-NR (approved):

ndesai-newrelic:
@cdillard-NewRelic the change is only required  for collector name.

PR #329, opened 2024-11-25
Changing auto collect logs to disabled by default

mbruin-NR:


cdillard-NewRelic (approved):
As long as tests do not need update. 👍

PR #338, opened 2025-01-08
Nr 350676 changing handled exceptions to send one per request

mbruin-NR:
This is the simplest patch to change handled exceptions to send one per request without rearchitecting. 
We are planning to rewrite this code so this PR may not be necessary but open to input.


libMobileAgent/src/Hex/src/HexPersistenceManager.cxx
```diff
@@ -17,32 +18,26 @@ HexPersistenceManager::HexPersistenceManager(std::shared_ptr<HexStore>& store,
...
+            // Create a new context for each piece of agent data
+            auto context = std::make_shared<HexContext>();
+            flatbuffers::Offset<HexAgentData> agentDataOffset = HexAgentData::Pack(*context->getBuilder(), agentDataObj.get(), nullptr);
 
-    if (agentDataVector.empty()) {
-     return nullptr;
-    }
-    auto bundle = fbs::CreateHexAgentDataBundle(*context->getBuilder(),
-                                             context->getBuilder()->CreateVector(agentDataVector));
+            auto agentDataVector = context->getBuilder()->CreateVector(&agentDataOffset, 1);
```
cdillard-NewRelic, line 31: Does it need to check this like the old code?
```
    if (agentDataVector.empty()) {
     return;
    }
```

libMobileAgent/src/Hex/src/HexPersistenceManager.cxx
```diff
@@ -17,32 +18,26 @@ HexPersistenceManager::HexPersistenceManager(std::shared_ptr<HexStore>& store,
...
+            auto bundle = fbs::CreateHexAgentDataBundle(*context->getBuilder(), agentDataVector);
+            FinishHexAgentDataBundleBuffer(*context->getBuilder(), bundle);
 
-    FinishHexAgentDataBundleBuffer(*context->getBuilder(), bundle);
+            // Publish the context for this agent data
+            _publisher->publish(context);
```
cdillard-NewRelic, line 36: Does it need to check that context != nil like the old code?

```
        auto context = _persistenceManager->retrieveStoreReports();
        if (context) {
            _publisher->publish(context);
        }
```

libMobileAgent/src/Hex/include/Hex/HexReportContext.hpp
```diff
@@ -7,14 +7,16 @@
 #define LIBMOBILEAGENT_HEXREPORTCONTEXT_HPP
 
 #include <Hex/HexContext.hpp>
+#include <Hex/HexPublisher.hpp>
 
 namespace NewRelic {
     namespace Hex {
-        class HexReportContext : public HexContext {
+        class HexReportContext : public HexContext, public std::enable_shared_from_this<HexReportContext> {
```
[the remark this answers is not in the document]
mbruin-NR: https://en.cppreference.com/w/cpp/memory/enable_shared_from_this
The std::enable_shared_from_this is used to safely obtain a std::shared_ptr to *this. This is what enables the ability to cast it as a HexContext and send it to the publisher.

HexReportContext::finalize() uses the shared_from_this() function to create a std::shared_ptr cast as a HexContext to send to publish.

libMobileAgent/src/Hex/src/HexPersistenceManager.cxx
mbruin-NR: Though the same name that variable is a different type now. It use to be an std::vector<flatbuffers::Offset<fbs::HexAgentData>> now it is a Offset<Vector<Offset<HexAgentData>>>.
The reports are read one at a time so that check is not needed. The std::vector is now a FlatBuffer Vector to match the data but I think it will always be one.

cdillard-NewRelic (approved):

smalsam-newr (approved):

PR #340, opened 2025-01-22
NR-359870 fix for the attribute validator crash

mbruin-NR:
Moved the attribute validator to utils so it can be static and used anywhere it's needed


smalsam-newr (changes requested):
The attribute validator doesn't need to be made static.

cdillard-NewRelic (dismissed):
Looks good to me 👍

cdillard-NewRelic:
I think this is okay use of static.
If there is concern with making this static I would suggest adding it as a singleton to the BlockAttributeValidator class.

smalsam-newr (commented):
This is not the way that class is meant to be used.

cdillard-NewRelic (approved):
LGTM

Agent/HandledException/NRMAHandledExceptions.mm
```diff
@@ -71,6 +74,8 @@ - (instancetype) initWithAnalyticsController:(NRMAAnalytics*)analytics
     self = [super init];
     if (self) {
         analyticsParent = analytics;
+        
+        _attributeValidator = [[NRMAAttributeValidator alloc] init];
```
smalsam-newr, line 78: If you pass something that conforms to the AttributeValidator protocol in as a parameter to this class, then you can replace it for testing.

smalsam-newr (approved):

Agent/HandledException/NRMAHandledExceptions.mm
mbruin-NR: Made the change to move it to be passed in.

smalsam-newr (approved):

cdillard-NewRelic (approved):

PR #342, opened 2025-01-29
NR-352010 Changed NRMASAM to add the persisted attributes individually so that they get re-persisted

mbruin-NR:
Changed NRMASAM to add the persisted attributes individually so that they get re-persisted which is closer to how the old event system does it in SessionAttributeManager::restorePersistentAttributes().
Now the session attributes should persist until removed like in the old event system.


smalsam-newr (approved):

cdillard-NewRelic (dismissed):
wanted to run your code locally and see what happens on bg, looks good 👍

cdillard-NewRelic (approved):
Good testing.

PR #343, opened 2025-01-30
Prevent setUserId nil log

cdillard-NewRelic:


Agent/Analytics/NRMASAM.mm
```diff
@@ -90,6 +90,10 @@ -  (BOOL) checkAttribute:(NSString*)name value:(id)value {
     BOOL validAttribute = [attributeValidator nameValidator:name];
     BOOL validValue = [attributeValidator valueValidator:value];
 
+    if (name == kNRMA_Attrib_userId && !value ) {
```
smalsam-newr, line 93: Do we want this to continue being an actual nil value, or do we want to put a placeholder value in for when it gets parsed on the backend?

smalsam-newr (dismissed):

Agent/Analytics/NRMASAM.mm
cdillard-NewRelic: This is a good question. It appears to me that userId is not a real "reserved" session attribute. Essentially an additional layer was added on the user attribute named "userId"

In our current impl we set the userId with
```
    BOOL success = [[NewRelicAgentInternal sharedInstance].analyticsController setSessionAttribute:kNRMA_Attrib_userId
                                                                                             value:userId
                                                                                        persistent:YES];
```
and if setting to nil we also call to remove

```
    // If passed userId == NULL , remove UserId attribute.
    if (userId == NULL) {
        success = [[NewRelicAgentInternal sharedInstance].analyticsController removeSessionAttributeNamed:kNRMA_Attrib_userId];
    }
```                                                                                        
                                                                                        

```
+ (BOOL) setUserId:(NSString* _Nullable)userId {
    NSString *previousUserId = [[NewRelicAgentInternal sharedInstance] getUserId];
    BOOL newSession = false;
    // If the client passes a new userId that is non NULL.
    if (userId != NULL) {
        // A new userId has been set where the previously set one (during this app session (since app launch)) was not NULL or the previous set one was NULL, we start a new session.
        newSession = true;
    }
    // If the client passes a new NULL userId.
    else {
        if (previousUserId != NULL) {
            // end session and harvest.
            newSession = true;
        }
        // Do nothing if passed userId is null and saved userId (for this app session (since app launch)) is null.
    }

    // Update in memory userId.
    [NewRelicAgentInternal sharedInstance].userId = userId;

    if (newSession) {
        [[[NewRelicAgentInternal sharedInstance] analyticsController] newSession];

        // Perform harvest
        [self harvestNow];

        [[NewRelicAgentInternal sharedInstance] sessionStartInitialization];
    }

    BOOL success = [[NewRelicAgentInternal sharedInstance].analyticsController setSessionAttribute:kNRMA_Attrib_userId
                                                                                             value:userId
                                                                                        persistent:YES];
    // If passed userId == NULL , remove UserId attribute.
    if (userId == NULL) {
        success = [[NewRelicAgentInternal sharedInstance].analyticsController removeSessionAttributeNamed:kNRMA_Attrib_userId];
    }

    return success;
}
````

PR #348, opened 2025-02-20
NR-347591 added a synchronized lock to the deferredMetrics

mbruin-NR:
These changes ensure that access to deferredMetrics is thread-safe. Multiple threads can modify deferredMetrics so synchronization blocks were added to protect access.


cdillard-NewRelic (approved):
Ran PR. Looks good.

cdillard-NewRelic (approved):

smalsam-newr (approved):

PR #356, opened 2025-03-25
NR-382686 Added thread safety to the tracked headers array

mbruin-NR:


cdillard-NewRelic (approved):
Lookin good as long as tests pass.

PR #370, opened 2025-04-22
main -> develop

mbruin-NR:


cdillard-NewRelic (approved):
changes look good. Approved pending test passing

PR #385, opened 2025-05-15
NR-378705 progress commit for image data

mbruin-NR:
Adding the ability to have image data in the session replay frame but left it commented out because it still needs some details worked out on what we can make work.


cdillard-NewRelic:
Seems okay. I lean the route of not committing things if we are just going to leave it all commented. lets just wait and implement once it works. If you want to add a bool that determines if images are used that sounds good to me.

cdillard-NewRelic (changes requested):
see comment

cdillard-NewRelic (approved):

smalsam-newr (approved):

PR #387, opened 2025-05-28
Rename NewRelic to NewRelicAgent

smalsam-newr:
This branch renames the main NewRelic class in order to avoid module conflicts with the NewRelic Module.


cdillard-NewRelic (approved):
👍 this is approved once tests compile,  run, and pass

PR #392, opened 2025-06-04
NR-415835 hasReplay attribute

mbruin-NR:
Added hasReplay to session attributes when session replay is recording. We add the attribute when the timers for session replay start. We remove the attribute when a new session starts, it will be added again if the session replay starts again.

### DIFF
--- a/Agent/SessionReplay/SessionReplayManager.swift
+++ b/Agent/SessionReplay/SessionReplayManager.swift
@@ -8,6 +8,7 @@
 
 import Foundation
 import UIKit
+import NewRelicPrivate
 
 @available(iOS 13.0, *)
 @objcMembers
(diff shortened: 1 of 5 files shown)


Agent/SessionReplay/SessionReplayManager.swift
```diff
@@ -36,7 +37,8 @@ public class SessionReplayManager: NSObject {
                 NRLOG_WARNING("Session replay harvest timer attempting to start while already running.")
                 return
             }
-            
+            NewRelicAgentInternal.sharedInstance().analyticsController.setNRSessionAttribute("hasReplay", value: NRMABool(bool: true))
```
smalsam-newr, line 40: You'll probably want to add a similar call for when we stop recording.

smalsam-newr (approved):

Agent/SessionReplay/SessionReplayManager.swift
mbruin-NR: I was wondering about that. Is there a case where a replay would start recording but then stop before the end of a session?

cdillard-NewRelic (approved):

PR #408, opened 2025-07-08
[NR-427117 ] feat: persist MSR

cdillard-NewRelic:
See NR-427117: https://new-relic.atlassian.net/browse/NR-427117

[NR-427117] feat: MSR persistence: initial checkForPreviousSessionFiles and processFrameToFile changes

---
As a customer, I want to save session Replay data during crashes, so I can see it on next upload.

In the event of a crash, much of the agent’s data is persisted to disk, to be harvested upon the next launch. Currently, the session replay data is not among the data saved and restored. We need to persist that to disk, load it upon the next launch, and upload it. This should be done in a manner similar to the other data which is persisted through a crash. Notably, we need to make sure the SessionID is persisted, so this last bit of replay data can be matched with the correct session

[NR-427117]: https://new-relic.atlassian.net/browse/NR-427117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

----

Testing Instructions:

1. Run application without debugger and cause a crash. When crash occurs go ahead and restarting app. You should see a request to the /blob endpoint including the frame data form the previous session.


Agent/SessionReplay/NRMASessionReplay.swift
```diff
@@ -116,15 +126,47 @@ public class NRMASessionReplay: NSObject {
...
+            if clear {
+                self.rawFrames.removeAll()
+
+                // should remove frames from file system after processing
+
+                // Clear the session replay file after processing
+                DispatchQueue.global(qos: .background).async { [self] in
+
+                    self.frameCounter = 0
+                    self.hasMeta = false
```
mbruin-NR, line 159:
```suggestion
```

Agent/SessionReplay/NRMASessionReplay.swift
```diff
@@ -159,15 +201,152 @@ public class NRMASessionReplay: NSObject {
...
+        let firstTimestamp: TimeInterval = TimeInterval(processedFrames.first?.timestamp ?? 0)
+        let lastTimestamp: TimeInterval = TimeInterval(processedFrames.last?.timestamp ?? 0)
+
+
+        var container: [AnyRRWebEvent] = []
+
+//        if frameCounter == 0 {
+            let metaEventData = RRWebMetaData(
+                href: "http://newrelic.com",
+                width: Int(self.windowDimensions.width),
```
mbruin-NR, line 236: The windowDimension variable is not updated because it is used to add a metaEvent to the start that contains the initial window size.
You should be able to use the frame to get the size for this. 
It should be fine to add a meta event for every frame, but is only needed before the first frame or at an orientation change.

Agent/SessionReplay/NRMASessionReplay.swift
```diff
@@ -159,15 +201,152 @@ public class NRMASessionReplay: NSObject {
...
+
+
+    func processFrameToFile(_ frame: SessionReplayFrame) {
+        // Fetch processed frame and touches during frame
+        let processedFrame = self.sessionReplayFrameProcessor.processFrame(frame)
+        let processedTouches = self.getSessionReplayTouches(clear: false)
+
+        let processedFrames = getSessionReplayFrames(clear: false)
+
+        let firstTimestamp: TimeInterval = TimeInterval(processedFrames.first?.timestamp ?? 0)
```
mbruin-NR, line 227: Is there a way to get these timestamps without processing all the frames every time?
The SessionReplayFrame has a Date variable.

cdillard-NewRelic: I don't thinnk we can, the purpose of this is to update the peristed URL to disk including the latest lastTimestamp and latestFirstTimestamp, can we get that from the one latest processed frame?

Agent/SessionReplay/NRMASessionReplay.swift
cdillard-NewRelic: Fixed.

Agent/SessionReplay/NRMASessionReplay.swift
```diff
@@ -159,15 +199,145 @@ public class NRMASessionReplay: NSObject {
...
     }
+
+
+    /// REPLAY PERSISTENCE
+
+
+    func processFrameToFile(_ frame: SessionReplayFrame) {
+        // Fetch processed frame and touches during frame
+        let processedFrame = self.sessionReplayFrameProcessor.processFrame(frame)
+        let processedTouches = self.getSessionReplayTouches(clear: false)
```
mbruin-NR, line 221: Processing and adding the same touches every frame probably won't cause issues in the playback but if possible it'd be nice to only record each touch once.

cdillard-NewRelic: 👍

Agent/SessionReplay/NRMASessionReplay.swift
cdillard-NewRelic: Changed to be exactly as you requested in a DM.

PR #409, opened 2025-07-09
[NR-411886] added touch masking check, changed isMasked logic to work better with defaults

mbruin-NR:
- If a touch starts in a masked view it will not be recorded no matter what.
- Account for if a unmasked touch moves into a masked view after LP.

- Changed the isMasked logic to account for the 4 different global masking bools.


cdillard-NewRelic (approved):
Excellent ctch fixing global mode. Initial touch tracking masking look good :+1:

PR #411, opened 2025-07-14
[NR-439820] fix for applicationId / accountId DT race condition 

mbruin-NR:
Save the accountId and and applicationId to be applied to the default configuration after it arrives from the server.


cdillard-NewRelic (approved):
As it was explained to me this fixes a scenario like backgrounding / foregrounding or maybe just setUserId 👍

cdillard-NewRelic (approved):
👍

PR #423, opened 2025-07-18
Added more custom masking / unmasking objects to test

mbruin-NR:


cdillard-NewRelic (approved):
A very good addition to TextMaskingVC for our sanity. Ty

PR #429, opened 2025-07-22
Added a check for contains .nr-mask or .nr-unmask

mbruin-NR:


cdillard-NewRelic (approved):

Agent/SessionReplay/ViewCaptors/ViewDetails.swift
```diff
@@ -85,7 +85,7 @@ struct ViewDetails {
     // This function checks if there are any specfic masking rules assigned to a view. If it returns nils, the masking value will be assigned based on the value of the global based on it's type later.
     private static func checkIsMasked(view: UIView, viewName: String) -> Bool? {
         // Determine if this view should be masked
-        var agent = NewRelicAgentInternal.sharedInstance()
```
cdillard-NewRelic, line 88: Is there a reason you had to make this change? I know I made NewRelicAgentInternal.sharedInstance() nullable to fix a crash. Is it giving a warning or something? If the change is required Im good wit it.

cdillard-NewRelic (approved):
👍

Agent/SessionReplay/ViewCaptors/ViewDetails.swift
cdillard-NewRelic: Ah can you please change this to 

```
        guard let agent = NewRelicAgentInternal.sharedInstance() else { return true }

cdillard-NewRelic (approved):

PR #430, opened 2025-07-22
Develop 7.5.7 -> staging

mbruin-NR:


Test Harness/NRTestApp/upload_dsym_results.log
```diff
@@ -0,0 +1,2 @@
+New Relic: Starting dSYM upload script...
```
cdillard-NewRelic, line 1: This shouldn't be committed and upload_dsym_results.log should be added to .gitignore

cdillard-NewRelic (approved):
👍

PR #436, opened 2025-07-24
Staging patch 3

mbruin-NR:


cdillard-NewRelic (approved):
Nice changes, only thing I see is that the ViewController added timeLabel needs to restart on app foregound , as to correspond with what MSR is actually doing (relative timestamp). That can be added in the next PR

PR #445, opened 2025-07-29
Changed NRMASAM to have a way of removing session attributes

mbruin-NR:


cdillard-NewRelic (approved):
I tested this in Prod for the enabled and disabled case and hasReplay is correct in these cases now.

PR #446, opened 2025-07-30
Adjusting some of the changes that were only for full frames to work also with diffs

mbruin-NR:


cdillard-NewRelic (approved):
Just wanted to confirm. These changes also effect full snapshot mode where diffs aren't used as well. I was testing a replay created with this branch and testing text entry

PR #453, opened 2025-08-07
[NR-445421] Added a supportability metric for URL too large response code

mbruin-NR:


Agent/SessionReplay/SessionReplayCapture.swift
```diff
@@ -12,7 +12,7 @@ import UIKit
 
 // potentially remove this annotation once the feature goes to Swift
 @objcMembers
-class SessionReplayCapture {
+internal class SessionReplayCapture {
```
cdillard-NewRelic, line 15: Do you know if this is required or if internal is the default?

cdillard-NewRelic (dismissed):
Excellent 👍

PR #464, opened 2025-08-18
Fix for the text of a UITextField not showing in the correct position

mbruin-NR:
This pr is an attempt to fix the text of a UITextField not showing in the correct position in the replay.
I decided to guess at where the text is by calculating it based an analyzing the text instead of searching the view tree for UITextLayoutFragmentView which is where the text actually should be. This decision is to reduce searching which has to occur on the main thread.

The if the text reaches the end of the view the text will be cut off. We could change it so that the text moves while it gets longer than the view but I don't think we could make it perfect. Like if you then scroll to the front of the text and add some text the text moves to the front.


cdillard-NewRelic (dismissed):
Awesome good addition to fix fidelity of replay 👍

cdillard-NewRelic (approved):

PR #523, opened 2025-10-30
NR-483348 

mbruin-NR:
On iPad a panel being added to the view could lose it's elements when using incremental diffs.
This pr changes it so that the number of UILayoutContainerViews in a UIPanelControllerContentView increasing can create a full snapshot.


cdillard-NewRelic (approved):
i ran this and ran tests against swiftui and uikit masking views and it appeared ok. I will try with sample apps Landmarks and sample app EnhancingTheAccessibilityOfYourSwiftUIApp

PR #557, opened 2025-12-02
Nr 488817 Fix for crash in UIColor.toHexString

mbruin-NR:
Changed the way UIColor.toHexString extracts colors and added a default of clear.


cdillard-NewRelic (approved):
👍

PR #560, opened 2025-12-11
Nr 500300 Fix for generateOptimizedPngData crash

mbruin-NR:
This prevents the crash by not attempting to create a context for zero-sized images.


cdillard-NewRelic (approved):
👍 well done

PR #691, opened 2026-04-15
Validate attribute's value types on NRMAAttributeValidator

fbernardo:
While using the "old event system", attributes passed to `[NRMAAnalytics addEventNamed:withAttributes:]` have their type checked for validation and aren't added to the event if invalid (done in `[NRMAAnalytics event:withAttributes:]`.

With the new event system validation moved to `NRMAAttributeValidator` that before this change allowed invalid types.
This would cause an issue where, during harvest an entire array of events would be discarded:

```objective-c
- (nullable NSString *)getEventJSONStringWithError:(NSError *__autoreleasing *)error clearEvents:(BOOL)clearEvents {
    NSString *eventJsonString = nil;
    @synchronized (events) {
        @try {
            NSMutableArray *jsonEvents = [[NSMutableArray alloc] init];
            for(id<NRMAAnalyticEventProtocol> event in events) {
                [jsonEvents addObject:[event JSONObject]];
            }
            
            NSData *eventJsonData = [NRMAJSON dataWithJSONObject:jsonEvents
                                                         options:0
                                                           error:error];
            eventJsonString = [[NSString alloc] initWithData:eventJsonData
                                                    encoding:NSUTF8StringEncoding];
            [self empty];
        } @catch (NSException *e) {
            NRLOG_AGENT_ERROR(@"FAILED TO CREATE EVENT JSON: %@", e.reason);
        }
    }
    if (clearEvents){
        [self empty];
    }
    return eventJsonString;
}
```

`eventJsonData` would be empty if the events failed to serialize, silently cleaning the array of events and sending nothing.

**This change introduces that validation missing to the validator and updates de tests to add that case.**


Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-551918

cdillard-NewRelic:
Thanks for the change. We intentionally made this less restrictive in the new event system. Thanks!

fbernardo:
@cdillard-NewRelic But then events (including ones with valid attributes) are being dropped during harvest. I guess that is not intentional. Is the intention to accept attribute values that aren't Strings, numbers or booleans?

cdillard-NewRelic:
Agreed it is confusing. I've taken this back to the team and we'll discuss merging your change and potentially adding better documentation about this case. Thank you!

cdillard-NewRelic:
Can you please attach the sample app or example code proving this issue out? Im unable to reproduce it.

cdillard-NewRelic:
Sorry but upon further review - for clarity can you please add the following to your PR?

- a button in NRTestApp that interacts with public facing NewRelic lib API to cause the issue

fbernardo:
I've included a new button on the NRTestApp. It logs two events, one valid, one invalid. Without this fix both return true.

At the harvest time (on `getEventJSONStringWithError`) ALL events are dropped (because empty is called) despite the JSON being nil due to encoding failing. **One invalid attribute on one event poisons the entire harvest.**

CleanShot 2026-06-11 at 16 59 55@2x

The intention of this change is to quickly return false when recording the event. Making the behaviour the same between the "old" and the new event system.

cdillard-NewRelic:
Missing the important setAttribute example. I've added it. Okay we can merge.

```
        options.append(UtilOption(title: "Set Invalid Type Attribute", handler: { [self] in setInvalidTypeAttribute()}))
```
```
    // Exercises the attribute value-type validation fix: a value whose type is not
    // NSString/NSNumber/BOOL (here an array) must be rejected. setAttribute returns
    // false after the fix (PASS); before the fix the invalid type was accepted (FAIL).
    func setInvalidTypeAttribute() {
        let accepted = NewRelic.setAttribute("invalidType", value: ["an", "array"])
        let result = accepted ? "FAIL: invalid-type attribute was accepted" : "PASS: invalid-type attribute was rejected"
        print("setInvalidTypeAttribute -> \(result)")
    }
```

cdillard-NewRelic (approved):

PR #696, opened 2026-04-17
NR-552783 add a breadcrumb when config changes like the Android agent

mbruin-NR:


cdillard-NewRelic:
While I approve of this, does this mean we are charging customers for a breadcrumb the agent adds?

cdillard-NewRelic (approved):

PR #708, opened 2026-04-24
NR-540532 added telemetry fallback if the zipped archive exceeds the 200MB threshold

mbruin-NR:


cdillard-NewRelic (approved):
Nice work. I've carefully reviewed this addition and verified it. Looks like this applies to both map and dsym endpoint 👍

PR #717, opened 2026-05-06
NR-556426 fix for hasReplay getting added to errorMode before fullMode is started

mbruin-NR:
- Not adding the hasReplay attribute in error mode before full mode is activated.
- Retroactively adding the hasReplay attribute to crashes when there is replay data.


cdillard-NewRelic (dismissed):
👍 well done

cdillard-NewRelic (approved):
👍

PR #728, opened 2026-05-13
NR-561596 NRFeatureFlag_JSErrorEvents feature flag added

mbruin-NR:
Also fixed a bug where MSR wouldn't work with crash reporting disabled.


cdillard-NewRelic (dismissed):
Excellent catch with MSR and crash reporting enabled. 👍

cdillard-NewRelic (approved):

cdillard-NewRelic (approved):

PR #754, opened 2026-05-29
NR-568934 fix: replace timestamp-seeded LCG in IGuidGenerator with arc4random_buf

diegomtz5:
## Summary
  - Fixes NR-568934 — iOS agent emits identical trace/span IDs across regional apps, producing fake cross-region
  service edges in Mobile → Dependencies. Reproduces against 7.6.3 and 7.7.1.
  - Root cause: `IGuidGenerator::generateGuid` reseeded `std::default_random_engine` per call from
  `system_clock::now()` truncated to 32 bits — same-tick calls produce identical seeds and identical output.
  - Fix: use `arc4random_buf` (kernel CSPRNG) for each ID. No shared seed, no same-tick collision class.

  ## Test plan
  - [x] `TestGuidUniqueness.mm` added; all 5 tests pass on the patched code.
  - [x] Pre-fix verification: with the LCG restored, `testFacadeNewPayloadTightLoopProducesNoDuplicates` reports
  26,454 duplicate span IDs in 100k calls (first duplicate at iter 20); concurrent variant reports 42,530 in 320k
  calls.
  - [ ] Reviewer: verify post-fix run shows 0 duplicates across all 5 tests.
  - [ ] Reviewer (independent collision repro): temporarily revert `IGuidGenerator.hpp` to the pre-PR LCG version on a
   scratch branch, ⇧⌘K (clean build), re-run `TestGuidUniqueness` — expect
  `testFacadeNewPayloadTightLoopProducesNoDuplicates` and
  `testFacadeNewPayloadTightLoopUnderConcurrencyProducesNoDuplicates` to fail with tens of thousands of duplicates.
  This independently confirms the test is exercising the right code path *and* that the fix is what's keeping it
  green.

  ## Customer comms
  Customer needs to upgrade to whichever release ships this fix or set the use the New Event System Feature Flag. Existing colliding spans in the backend will age out
  as the 75-min span TTL elapses, but only once *all* regional apps are on the patched build — mixing patched and
  unpatched apps in the fleet leaves the collision class active.


cdillard-NewRelic (approved):
:+1: well done

PR #758, opened 2026-06-01
fix: NR-560574 recursion blockage for maui style default interaction swizzle

cdillard-NewRelic:
https://new-relic.atlassian.net/browse/NR-560574

 The Xamarin/MAUI infinite-recursion crash fix is implemented and builds. Files changed:

  - NRMAActingClassUtils.h / .m — added NRMA_enterSelector / NRMA_exitSelector, a per-thread, per-(self, selector) re-entry depth counter.
  - NRMAMethodProfiler.m — NRMA__beginMethod now forces the super-advance when the method color matches but we're re-entering the same (self, 
  selector) at depth > 1 (the non-advancing pattern Xamarin's xamarin_dyn_objc_msgSendSuper causes); NRMA__endMethod and the @throw path
  balance the counter.
  
  The recursion now terminates at the parent's real implementation instead of overflowing the stack, and legitimate [super _cmd] calls are
  unaffected because they produce a genuine color mismatch (the && isTargetColor guard never fires for them).


cdillard-NewRelic (commented):
Approved.

diegomtz5 (approved):

PR #764, opened 2026-06-04
NR-376961 Added a supportability metric to count KMP users

mbruin-NR:
Adds automatic detection of Kotlin Multiplatform (KMP) frameworks in iOS applications and reports usage via a supportability metric. This enables New Relic to track KMP adoption and provide better support for Kotlin-based hybrid apps.

  Changes Made

  Core Detection Logic (Agent/Utilities/NewRelicInternalUtils.m)
  - Added detectKMMFrameworks method that scans the Objective-C runtime for Kotlin/Native classes
  - Uses objc_getClassList to enumerate all registered classes at runtime
  - Detects Kotlin frameworks by identifying classes containing "Kotlin" in their names (e.g., SharedKotlinEnum, MyAppKotlinArray, KotlinBase)

  Supportability Metric (Agent/Measurements/NRMASupportMetricHelper.m)
  - Added enqueueKMMDetectionMetric to report KMP detection
  - Metric name: Supportability/Mobile/iOS/HybridPlatform/KMP
  - Sent once per app session when KMP frameworks are detected

  Technical Implementation

  Detection Strategy:
  The implementation scans all Objective-C classes registered in the runtime. Kotlin/Native generates Objective-C-compatible classes with predictable naming patterns:
  - Framework classes: {FrameworkName}Kotlin{ClassName} (e.g., SharedKotlinArray)
  - Standard library: KotlinBase, KotlinEnum, KotlinException, etc.

  Safety Features:
  1. Integer overflow protection - Guards against unreasonably large class counts (>100,000)
  2. Race condition handling - Allocates extra buffer space (128 classes) to handle classes registered between API calls
  3. Buffer overflow prevention - Always iterates within allocated buffer bounds
  4. Memory safety - Validates malloc, checks NULL pointers, ensures proper cleanup on all code paths
  5. Defensive coding - Validates string conversion before use

  Why This Approach?

  Alternative approaches considered:
  - ❌ Symbol lookup (dlsym) - Unreliable due to symbol visibility and underscore prefixes
  - ❌ dyld image scanning - More complex, requires framework name patterns
  - ✅ Runtime class enumeration - Simple, reliable, works regardless of framework name

### DIFF
--- a/Agent/General/NewRelicAgentInternal.m
+++ b/Agent/General/NewRelicAgentInternal.m
@@ -320,6 +320,12 @@ static NewRelicAgentInternal* _sharedInstance;
     // Update old files (no more files in the docs folder)
     [NRMAFileCleanup updateDocFileLocations];
 
+    // Detect KMP (Kotlin Multiplatform) frameworks for supportability metrics
+    BOOL kmpDetected = [NewRelicInternalUtils detectKMMFrameworks];
+    if (kmpDetected) {
+        [NRMASupportMetricHelper enqueueKMMDetectionMetric];
+    }
+
     NRMAExceptionHandlerStartupManager* exceptionHandlerStartupManager = [[NRMAExceptionHandlerStartupManager alloc] init];
 
     // Last session's analytics must be fetched (asynchronously) before instrumentation
--- a/Agent/Measurements/NRMASupportMetricHelper.h
+++ b/Agent/Measurements/NRMASupportMetricHelper.h
@@ -41,4 +41,6 @@ static NSMutableArray *deferredMetrics;
 + (void) enqueueJSErrorUploadThrottledMetric;
 + (void) enqueueJSErrorFailedUploadMetric;
 
++ (void) enqueueKMMDetectionMetric;
+
 @end
--- a/Agent/Measurements/NRMASupportMetricHelper.m
+++ b/Agent/Measurements/NRMASupportMetricHelper.m
@@ -278,6 +278,19 @@ static NSMutableArray<NRMAMetric *> *deferredMetrics;
 
 // End JS Error
 
+// KMP Detection (Kotlin Multiplatform)
++ (void) enqueueKMMDetectionMetric {
+    @synchronized (deferredMetrics) {
+        [deferredMetrics addObject:[[NRMAMetric alloc] initWithName:kNRMAKMMDetectionMetric
+                                                              value:@1
+                                                              scope:@""
+                                                    produceUnscoped:YES
+                                                    additionalValue:nil]];
+    }
+}
+
+// End KMP Detection
+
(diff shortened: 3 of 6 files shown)


Agent/General/NewRelicAgentInternal.m
```diff
@@ -320,6 +320,12 @@ - (void) initialize {
     // Update old files (no more files in the docs folder)
     [NRMAFileCleanup updateDocFileLocations];
 
+    // Detect KMP (Kotlin Multiplatform) frameworks for supportability metrics
+    BOOL kmpDetected = [NewRelicInternalUtils detectKMMFrameworks];
```
cdillard-NewRelic, line 324: What thread does objc_getClassList get called on? Does it matter? I see that its a new usage in the agent.
I see in NRMAMethodProfiler.m we use 
```
NSMutableDictionary* NRMA__generateClassTrees(NSSet* parents)
{
    const char *mainImageName = [[[NSBundle mainBundle] executablePath] fileSystemRepresentation];
    NSMutableDictionary* results = [[[NSMutableDictionary alloc] init] autorelease];

    unsigned int classNameCount = 0;
    const char **classNames = objc_copyClassNamesForImage(mainImageName, &classNameCount);
    for (NSUInteger i = 0; i < classNameCount; i++) {
        Class cls = objc_getClass(classNames[i]);
        NRMA__processClass(cls, results, parents);
    }
    free(classNames);

    for (NSString *className in [NRMAMethodProfiler whiteList]) {
        Class cls = NSClassFromString(className);
        if (cls) {
            NRMA__processClass(cls, results, parents);
        }
    }

    return results;
}
```
Just curious if that is the same or preferred?

mbruin-NR: Getting onto another thread is probably a good idea.
  
NRMA__generateClassTrees (line 425-446):
  const char *mainImageName = [[[NSBundle mainBundle] executablePath] fileSystemRepresentation];
  const char **classNames = objc_copyClassNamesForImage(mainImageName, &classNameCount);
  - Uses objc_copyClassNamesForImage
  - Only scans classes from one specific image (the main app bundle)
  - Purpose: Find app's own classes for method profiling/swizzling

 detectKMMFrameworks implementation:
  int actualCount = objc_getClassList(classes, bufferSize);
  - Uses objc_getClassList
  - Scans ALL classes from the entire runtime (app + all loaded frameworks)
  - Purpose: Find Kotlin classes from dynamically linked frameworks

cdillard-NewRelic: 👍 Looks good thank you.

diegomtz5 (approved):

cdillard-NewRelic (approved):
Approved. Discussed performance and this is fine.

Issue #772, opened 2026-06-18
Distributed Tracing: injected W3C traceparent trace-id does not match the recorded network span's trace.id on the NSURLSession delegate path (breaks mobile -> backend trace linking)


mattk1ng:
## Summary

On the `NSURLSession` **delegate** code path (no completion handler), the agent injects a W3C `traceparent` header built from one payload, but records the corresponding network-request span (`addNetworkRequestEvent`) from a **different** payload with a **different** `trace.id`. As a result the outbound `traceparent` and the recorded mobile span never share a trace id, so a downstream service that correctly continues the inbound `traceparent` ends up in a separate trace from the mobile span. Mobile and backend never link.

This is the path React Native (`RCTNetworking`) and any delegate-based `NSURLSession` usage take, so it affects `newrelic-react-native-agent` users doing mobile -> backend distributed tracing.

There is also a secondary, related defect: the generated trace ids are low-entropy and frequently have two identical 64-bit halves.

## Environment

- `NewRelicAgent` (iOS) **7.7.2** (latest published), arm64 simulator + device
- Consumed via `newrelic-react-native-agent` **1.8.4**
- New event system enabled (`shouldEnableNewEventSystem`), `distributedTracingEnabled: true`
- Backend continues the inbound W3C `traceparent` via OpenTelemetry (OTLP), `ALWAYS_ON` sampler

## Impact

Mobile spans and the backend spans they trigger are placed in different traces, so mobile-to-backend distributed traces cannot be assembled. The backend faithfully continues a `traceparent` that has no mobile span behind it (an orphan trace).

## Evidence (real values, infra details genericized)

For a single `GET https://api.example.com/jobs`:

| Source | trace.id | span/parent id |
| --- | --- | --- |
| Recorded mobile span (`Span`, mobile entity) | `ea92f7bac8d41962f65be0e6b9c657ef` | id `0224d212aaf8967c` |
| Backend server span (continued the inbound header) | `7b2961adf3fdf1cd7b2961adf3fdf1cd` | parent.id `810dda43ec771114` (remote) |

NRQL `FROM Span WHERE trace.id LIKE '%7b2961adf3fdf1cd%'` returns only the backend spans - no mobile span exists under the injected id anywhere in the account. The backend span has a valid *remote* parent.id, confirming it genuinely continued an inbound `traceparent`; that id is simply not the one the mobile span was recorded under. This reproduces on every request.

[...]

PR #773, opened 2026-06-18
Distributed Tracing: align injected traceparent with recorded span trace.id on the setState path (#772)

mattk1ng:
Fixes the distributed-tracing payload-overwrite defect analyzed in #772 on the `NSURLSession` task `setState` path. This is the path instrumented when the `SwiftAsyncURLSessionSupport` feature flag is enabled.

> The trace-id entropy fix (`IGuidGenerator`) that was originally part of this PR has been split out into #775 at the maintainer's request, since it is already addressed on `develop` (PR #754, shipping in 7.7.3).

## The defect

`NRMAOverride__urlSessionTask_SetState` regenerated a DT payload and re-attached it to `task.originalRequest`, **overwriting** the payload attached at task creation (`NRMAURLSessionOverride`) whose `traceparent` was already injected onto the wire. The recording path (`NRMAURLSessionTaskDelegateBase` -> `retrieveNRMAPayload` -> `addNetworkRequestEvent`) then read the newer payload, so the recorded span's `trace.id` differed from the header the backend continues. It also generated two payloads (`addConnectivityHeaderAndPayload` + `addConnectivityHeaderNRMAPayload`), the first on a throwaway request that was discarded.

Fix: only generate/attach a payload here when one is **not already attached** (new non-destructive `+[NRMAHTTPUtilities hasAttachedPayload:]`, covering both event systems), and drop the redundant second generation.

## Scope note

This override is only swizzled in when `SwiftAsyncURLSessionSupport` is enabled (`NRMAURLSessionTaskOverride.m`: `if (![NRMAFlags shouldEnableSwiftAsyncURLSessionSupport]) return;`), and it early-returns for tasks the session-level swizzle already handled (`NRMAHandledRequestKey`). So this change only affects callers running with that flag on, on the Swift async / non-handled task path.

## Testing notes

- Minimal, targeted change. I was **not able to run the agent's full Xcode test harness locally**, so please validate against the `NSURLSession` instrumentation tests - in particular the delegate vs completion-handler task paths and the `isHandled` early-return interaction.
- Happy to adjust the approach - e.g. if you'd prefer the dedupe to live in the recording path (`NRMANetworkFacade`) rather than the task override.

Refs newrelic/newrelic-ios-agent#772

### DIFF
--- a/Agent/Instrumentation/NSURLSession/NRMAURLSessionTaskOverride.m
+++ b/Agent/Instrumentation/NSURLSession/NRMAURLSessionTaskOverride.m
@@ -134,19 +134,26 @@ void NRMAOverride__urlSessionTask_SetState(NSURLSessionTask* task, SEL _cmd, NSU
                 NSURL *url = [currentRequest URL];
                 if (url != nil) {
 
-                    // Added this section to add Distributed Tracing traceId\trace.id, guid,id and payload.
-                    //1
-                    NSMutableURLRequest* mutableRequest = [NRMAHTTPUtilities addCrossProcessIdentifier:currentRequest];
-                    mutableRequest = [NRMAHTTPUtilities addConnectivityHeaderAndPayload:mutableRequest];
-                    // 2
-                    if([NRMAFlags shouldEnableNewEventSystem]){
-                        NRMAPayload* payload = [NRMAHTTPUtilities addConnectivityHeaderNRMAPayload:mutableRequest];
-                        [NRMAHTTPUtilities attachNRMAPayload:payload
-                                                      to:task.originalRequest];
-                    } else {
-                        NRMAPayloadContainer* payload = [NRMAHTTPUtilities addConnectivityHeader:mutableRequest];
-                        [NRMAHTTPUtilities attachPayload:payload
-                                                      to:task.originalRequest];
+                    // Attach a distributed tracing payload for the recorded span ONLY if an
+                    // earlier instrumentation point (e.g. the NSURLSession-level override at
+                    // task creation) has not already attached one. Regenerating here used to
+                    // overwrite that payload - whose traceparent was already injected onto the
+                    // wire - so the recorded span's trace.id no longer matched the header the
+                    // backend continues, breaking mobile -> backend trace linking
+                    // (newrelic/newrelic-ios-agent#772). This also drops the redundant second
+                    // payload generation (addConnectivityHeaderAndPayload created a payload on a
+                    // throwaway request that was immediately discarded).
+                    if (![NRMAHTTPUtilities hasAttachedPayload:task.originalRequest]) {
+                        NSMutableURLRequest* mutableRequest = [NRMAHTTPUtilities addCrossProcessIdentifier:currentRequest];
+                        if([NRMAFlags shouldEnableNewEventSystem]){
+                            NRMAPayload* payload = [NRMAHTTPUtilities addConnectivityHeaderNRMAPayload:mutableRequest];
+                            [NRMAHTTPUtilities attachNRMAPayload:payload
+                                                          to:task.originalRequest];
+                        } else {
+                            NRMAPayloadContainer* payload = [NRMAHTTPUtilities addConnectivityHeader:mutableRequest];
+                            [NRMAHTTPUtilities attachPayload:payload
+                                                          to:task.originalRequest];
+                        }
                     }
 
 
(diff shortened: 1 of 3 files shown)


Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-578551

cdillard-NewRelic:
Thank you @mattk1ng I will take this back to the team and get back to you at a later date.

cdillard-NewRelic:
@mattk1ng Can you please sign our CLA agreement as mentioned above? Thanks!

mattk1ng:
> Thank you @mattk1ng I will take this back to the team and get back to you at a later date.

Thank you, I have signed the CLA agreement too.

cdillard-NewRelic:
@mattk1ng Can you please move the guid generator changes to a separate PR? We've addressed that in our develop branch already and it will be fixed in 7.7.3 release. Can you confirm if you are using the SwiftAsyncURLSessionSupport feature flag? If so your new code for 2) fixes it and we can merge it. Thanks.

mattk1ng:
## Withdrawing this PR - root cause is outside the agent (Expo dev-client artifact)

Thanks for engaging on this, and apologies for the noise. After a deeper investigation with a proper runtime repro, **this is not a New Relic agent bug**, and the fix in this PR is not the right one. Closing this PR (#773), the split-out entropy PR (#775), and the underlying issue (#772). Writing up the findings in full in case they are useful to you or to other React Native / Expo users.

### TL;DR
The agent is correct: on the `NSURLSession` delegate path it records each task's span with the **same** `trace.id` it injects on that task's `traceparent`. The orphan traces we reported are caused by **Expo's dev-client network inspector** (`expo-dev-launcher`), which re-issues every request on a second `URLSession` in **debug builds only**. The agent faithfully instruments both tasks; one is recorded, the other is the one that reaches the wire, so they appear to diverge. **Release/production builds do not have this and link correctly** (verified end to end).

### How we got there
Running a diagnostic build of the agent (7.7.2) and logging, per request, the `traceparent` injected vs the `trace.id` recorded:

- On the **stock** agent, every single task is internally consistent: `reqTraceparent == recordedTraceId` for `/auth/me`, `/jobs`, `/messages`, `/notification-preferences`. So the recording path is sound.
- But each request produced **two** `dataTaskWithRequest:` calls, on **two different `URLSession` objects**, while the JS `fetch` was called exactly **once**. `beginInstrumentation` ran once (no double-swizzle).

### The actual cause
The second `URLSession` is created by Expo's dev tools network inspector:

- `ExpoRequestInterceptorProtocol` (`expo-modules-core`, `ios/DevTools/`) is a `URLProtocol` with its **own** `URLSession` that re-performs the request to feed the Chrome-DevTools-Protocol "Network" tab.
- It is registered by `expo-dev-launcher` (`DevLauncherNetworkInterceptor.swift`), which swizzles `URLSessionConfiguration.default` to insert the protocol, and that whole class is gated behind **`#if DEBUG && EX_DEV_CLIENT_NETWORK_INSPECTOR`** (the non-DEBUG path is a no-op stub).

So in a debug/dev-client build, one app `fetch` becomes two `NSURLSession` tasks. The agent injects an independent DT payload into each (correctly, since they are two distinct tasks), records the span for one, while the interceptor's task is the one that actually hits the network. Result: the backend continues a `traceparent` whose mobile span was recorded under a different id -> orphan.

### Production behavior (verified end to end)
Built a Release-config app (`expo run:ios --configuration Release`), signed in, DT enabled. The recorded mobile spans match the backend-continued trace ids:

| Endpoint | Recorded mobile span `trace.id` | Backend-continued `trace.id` | |
| --- | --- | --- | --- |
| `GET /jobs` | `ed012a37e526969ef8ca1b63d618d52b` | `ed012a37e526969ef8ca1b63d618d52b` | linked |
| `PUT /notification-preferences` | `00f61a5ffe299a9d06da92f5f6a2b9e4` | `00f61a5ffe299a9d06da92f5f6a2b9e4` | linked |

These are the same two endpoints that orphaned in the dev build. In Release (where the interceptor compiles out), mobile and backend share the trace id and link correctly.

### Net
- The agent's NSURLSession instrumentation is behaving correctly; no change is needed here.
- The trace-id entropy point (#775) is real but cosmetic, and you have already addressed it on `develop` for 7.7.3.
- #772 is not an agent defect; it is the Expo dev-client network inspector double-issuing requests in debug builds.

Thanks again for the help, and sorry for the detour.

PR #774, opened 2026-06-18
NR-572211 fix infinite loop in EventDeserializer that froze the agent on launch

mbruin-NR:
## Summary

Fixes an app-launch freeze on React Native apps caused by an infinite loop
in `EventDeserializer` when the events dup store contains a record that
ends with consecutive `\t` delimiters. The agent's main thread would peg a
core during `AnalyticsCever return.

## Root cause

`AttributeDeserializer: the attribute name via
`is.get(streambuf&, '\t')`. Per the standard, when the very next character
already is the delimiteters and sets
`failbit` — but **not** `eofbit`. The per-attribute loops in
`EventDeserializer::des
`deserializeMobileEvent` only checked `!is.eof()`, so once the stream
landed in the failbit-over.

This fired in production empty attribute name
(yielding `\t\t` in the stream) or on partially-corrupted persisted data.

## Fix

Two complementary changes:

1. **Read side** (`EventDeserializer.cxx`): both per-attribute loops now
   exit on `is.fail()` s the load-bearing
   fix — it terminates the loop on any stream-state hazard, including
   data already on diskture corruption.
2. **Write side** (`AnalyticEvent.cxx`): `insertAttribute` rejects empty
   attribute names. `Atraw `\t`/`\n` in
   names at construction, so empty-name was the only remaining write-side
   path that could prodtate.

### DIFF
--- a/libMobileAgent/src/Analytics/include/Analytics/Stores/FileBackedStore.hpp
+++ b/libMobileAgent/src/Analytics/include/Analytics/Stores/FileBackedStore.hpp
@@ -5,10 +5,7 @@
 #include <Utilities/libLogger.hpp>
 #include <Utilities/WorkQueue.hpp>
 #include <Analytics/AnalyticEvent.hpp>
-#include <atomic>
 #include <chrono>
-#include <future>
-#include <memory>
 #include <sstream>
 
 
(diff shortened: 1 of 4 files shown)


cdillard-NewRelic:
That's pretty crazy that bad of bug was living in the C++ THis looks great though. I will approve.

cdillard-NewRelic (approved):
👍

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-578980

mbruin-NR:
> PersistentStore
I think it isn't already wired up to get the C++ libraries into our objective c unit tests.

cdillard-NewRelic:
I noticed that too when I tried to compile the tests. I'm OK without the test, no problem.

PR #781, opened 2026-06-23
fix: NRMAWKWebViewNavigationDelegate webViewWebContentProcessDidTerminate crash fix

cdillard-NewRelic:
…

Root cause

NRMAWKWebViewNavigationDelegate is a proxy the agent slots in front of the app's real WKNavigationDelegate. It holds the real delegate weakly (@property(weak) NSObject* realDelegate; — uniquely among the agent's delegate proxies; the NSURLSession/NSURLConnection ones retain theirs, which is why only WebKit crashes). The crash is a three-step race:

1. At delegate-assignment time the real delegate is alive and implements webViewWebContentProcessDidTerminate:, so the proxy's respondsToSelector: returns YES — WebKit caches this.
2. The owning controller deallocates → the weak realDelegate zeroes to nil.
3. The web view's content process is reclaimed (common when backgrounded) → WebKit fires the cached callback. The proxy doesn't implement it, so forwardingTargetForSelector: returns the now-nil delegate, the runtime falls through to methodSignatureForSelector: (returns nil) → doesNotRecognizeSelector: → NSInvalidArgumentException.

This is general to any optional WKNavigationDelegate method the base class doesn't implement directly — webViewWebContentProcessDidTerminate: is just the one that reliably fires after teardown.

The fix (NRMAWKWebViewNavigationDelegate.m + NRWKNavigationDelegateBase.m)

Hardened the proxy's message-forwarding layer so a vanished weak delegate is absorbed, not crashed — keeping the weak semantics (changing 's memory behavior / risk leaks):
- forwardingTargetForSelector: — fast-forward only when the delegate is alive and responds; else fall through.
- methodSignatureForSelector: / forwardInvocation: — when the delegate is gone, return a benign       signature and silently drop instead oector:.
- Defense-in-depth in invokeMethod: — snapshot the weak ref and bail if gone (same root cause; otherwi+invocationWithMethodSignature:nil th
                                                                                                      Verification
                                                                                                      - Reproduced the exact crash against solated clang harness:NSInvalidArgumentException: -[NRMAWKWebViewNavigationDelegate webViewWebContentProcessDidTerminate:]: unrecognized selector — matching the y.
- Red → green confirmed: after the fix, the deallocated-delegate scenario passes (no exception) and the
live-delegate scenario still forwards
- xcodebuild build -scheme Agent-iOS → ** BUILD SUCCEEDED ** with both files compiling cleanly.
- Added two unit tests in NRMAWKNaviging both scenarios.

### DIFF
--- a/Agent/Instrumentation/WKWebView/NRMAWKWebViewNavigationDelegate.m
+++ b/Agent/Instrumentation/WKWebView/NRMAWKWebViewNavigationDelegate.m
@@ -24,7 +24,44 @@
 }
 
 - (id)forwardingTargetForSelector:(SEL)aSelector {
-    return self.realDelegate;
+    // realDelegate is held weakly. WebKit caches the result of respondsToSelector:
+    // at navigation-delegate-assignment time, so it can later send us a selector that
+    // only the real delegate implements (e.g. webViewWebContentProcessDidTerminate:,
+    // which fires when a backgrounded web view's content process is reclaimed) after
+    // that real delegate has been deallocated. Only fast-forward when the real delegate
+    // is still alive and actually implements the selector; otherwise return nil so the
+    // message drops into the full forwarding machinery below and is absorbed rather than
+    // falling through to doesNotRecognizeSelector: and crashing the host app.
+    NSObject* delegate = self.realDelegate;
+    if ([delegate respondsToSelector:aSelector]) {
+        return delegate;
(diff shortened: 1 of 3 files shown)


diegomtz5 (approved):
Was just going through this, thanks! came to the same solution. Just ran it and it fixes it.

PR #786, opened 2026-06-25
NR-577140: Add TTL eviction for offline storage

cdillard-NewRelic:
1. Configurable TTL (default 7 days). Added kNRMADefaultOfflineStorageTTLSeconds (7 days, aligned with the shortest retention window — Session Replay ~8 days) plus class methods +defaultOfflineStorageTTL, +offlineStorageTTL, +setOfflineStorageTTL:. Kept it process-wide/static to match Android's static offlineStorageTTL.

2. TTL expiry on read (getAllOfflineData:). Before reading each file, it checks the file's NSFileModificationDate against now − TTL. Stale files are deleted (and the tracked size counter decremented) and skipped instead of being uploaded — so data past retention never gets re-uploaded, even when clear:NO.

3. LRU eviction on persist (persistDataToDisk:). When the incoming payload would exceed the cap, it now sorts existing files oldest-first and evicts them until the new payload fits, rather than dropping the current harvest. It only returns NO if eviction still can't free enough space (e.g., a single payload larger than the whole cap). Added helpers evictOldestFilesToFit: and removeOfflineFileAtPath:size:.

[SessionReplayReporterOfflineStorageTests.swift􀰓](file:///Users/cdillard/Documents/Workspace/newrelic-ios-agent/Tests/Unit-Tests/NewRelicAgentTests/Uncategorized/SessionReplayReporterOfflineStorageTests.swift) — added NRMAOfflineStorageTTLTests with 7 tests matching the Android PR's coverage: default TTL, set/get TTL, expired-deleted-on-read, non-expired-returned, mixed expired/valid, LRU evicts oldest, and persist-returns-false-when-eviction-can't-help.

Differences: 

• Android orders/expires by File.lastModified(); iOS uses the same concept via NSFileModificationDate.
• iOS tracks total size in a NSUserDefaults counter (rather than rescanning disk like Android's getTotalFileSize()), so eviction/expiry decrement that counter to keep it consistent (clamped at 0).
• One API difference worth flagging: iOS setMaxOfflineStorageSize: takes megabytes (×1,000,000), whereas Android's setOfflineStorageSize takes raw bytes — that's why the LRU tests use ~400 KB payloads against a 1 MB cap. I left that public API as-is to avoid scope creep.

### DIFF
--- a/Agent/Utilities/NRMAOfflineStorage.h
+++ b/Agent/Utilities/NRMAOfflineStorage.h
@@ -19,4 +19,11 @@
 - (NSString*) offlineDirectoryPath;
 + (NSString*) allOfflineDirectorysPath;
 
+// Time-to-live (in seconds) after which a persisted offline payload is considered
+// stale and is deleted instead of re-uploaded. Defaults to +defaultOfflineStorageTTL
+// (7 days), aligned with the shortest applicable backend retention window.
++ (NSTimeInterval) defaultOfflineStorageTTL;
++ (NSTimeInterval) offlineStorageTTL;
++ (void) setOfflineStorageTTL:(NSTimeInterval) seconds;
+
 @end
--- a/Agent/Utilities/NRMAOfflineStorage.m
+++ b/Agent/Utilities/NRMAOfflineStorage.m
@@ -15,8 +15,6 @@
 #import "NRMASupportMetricHelper.h"
 #import "NRMAAgentConfiguration.h"
 
-#define kNRMAOfflineStorageCurrentSizeKey @"com.newrelic.offlineStorageCurrentSize"
-
 // Default time-to-live for persisted offline payloads: 7 days, in seconds.
 // Aligned with the shortest applicable New Relic data-retention window (Session Replay ~8 days).
 #define kNRMADefaultOfflineStorageTTLSeconds (7 * 24 * 60 * 60)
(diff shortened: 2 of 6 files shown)


diegomtz5:
@cdillard-NewRelic found some issues going over this, not introduced by this but we might want to double check it, just pasting Claude's reply. I can go into more detail tomorrow and see if its accurate. 

Two offline-storage issues worth checking, both pre-existing on develop, not introduced by this PR, but the PR's counter-driven eviction depends on the size counter being accurate

1. Same-second filename collisions. newOfflineFilePath is second-resolution (yyyy-MM-dd-HH-mm-ss.txt). In sendOfflineStorage's replay loop, when the device is still offline every send: fails instantly and re-persists in the same second → files overwrite each other (buffered payloads lost) and the size counter over-counts. (The old testOfflineStorageMaxSize had a sleep(1) to dodge exactly this.)
2. Single global size counter shared across stores. com.newrelic.offlineStorageCurrentSize is shared by the data and sessionreplay stores, but each clearAllOfflineFiles deletes only its own dir while zeroing the shared counter. So either store's flush wipes the other's accounting → setMaxOfflineStorageSize drifts/becomes unenforceable.

Net: offline-storage size accounting can't be fully trusted from two independent angles, and #786 builds eviction on top of it. Probably worth a ticket each and deciding whether to address before/with this PR. Fix direction that covers both: compute on-disk size on demand instead of caching a counter.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-582005

diegomtz5 (dismissed):

cdillard-NewRelic:
THank you for that critical and thoughtful comment im going to double check the concerns and report back.

cdillard-NewRelic:
Addressed the "Net" of the review — "compute on-disk size on demand instead of caching a counter" — in Agent/Utilities/NRMAOfflineStorage.m. The shared NSUserDefaults counter com.newrelic.offlineStorageCurrentSize is gone entirely, replaced by a new private helper:

```
- (NSUInteger) currentOfflineStorageSize {
    // sums NSFileSize of every file in *this store's own* offline directory
}
```


Call sites updated:
• persistDataToDisk: — reads size from disk before and after eviction instead of from the counter.
• evictOldestFilesToFit: — seeds the running total from disk, then decrements locally per eviction (avoids re-scanning the directory each iteration, no O(n²)).
• removeOfflineFileAtPath: — dropped the now-pointless size: parameter; it just deletes the file.
• clearAllOfflineFiles / clearAllOfflineDirectories — no longer zero a shared counter.

Also switched the arithmetic back to NSUInteger (the old NSInteger casts only existed to guard a counter that could go negative; an on-disk sum can't).

Why this fixes both angles the review flagged
• Same-second filename collisions over-counting — when a collided write overwrites a file, the on-disk sum reflects reality automatically; there's no stale increment to drift.
• Single global counter shared across the data + sessionreplay stores — each store now sums its own directory, so one store's clearAllOfflineFiles can no longer corrupt the other's accounting, and setMaxOfflineStorageSize stays enforceable. The counter-driven eviction in #786 now sits on a trustworthy number.

cdillard-NewRelic:
The UUID makes uniqueness guaranteed (millisecond precision alone isn't enough in a tight synchronous loop on fast hardware). I verified nothing parses the old yyyy-MM-dd-HH-mm-ss.txt format — eviction and TTL both order by NSFileModificationDate, not the filename — so changing the format is safe.

Regression test — [NRMAOfflineStorageTests.m􀰓](file:///Users/cdillard/Documents/Workspace/newrelic-ios-agent/Tests/Unit-Tests/NewRelicAgentTests/Uncategorized/NRMAOfflineStorageTests.m), testRapidPersistsDoNotOverwriteEachOther: persists 5 payloads back-to-back (same second, mimicking the replay loop), then asserts 5 distinct files exist on disk, the on-disk size sums all 5, and all 5 are recoverable via getAllOfflineData. Before the fix this would have collapsed to 1 file.

diegomtz5 (approved):
Looks good!

mbruin-NR (approved):

PR #795, opened 2026-07-01
NR-581396 working per report delete from upload results

mbruin-NR:
Changed the hex uploader to track report uploads with a reportId so that they can be deleted when uploading either succeeds or fails three times.

### DIFF
--- a/Agent/HandledException/HexUploadPublisher.hpp
+++ b/Agent/HandledException/HexUploadPublisher.hpp
@@ -7,6 +7,7 @@
 #define NEWRELICAGENT_HEXUPLOADPUBLISHER_H
 
 #include <Hex/HexPublisher.hpp>
+#include <functional>
 /*
  * Follows the HexPublisher interface for dependency injection into the NewRelic::Hex::HexController
  * This class uses a PIMPL to allow for Obj-c code to be called from C++ allowing access to the high-level networking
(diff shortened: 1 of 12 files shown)


cdillard-NewRelic (dismissed):
Claude has 3 qs, it looks good to me though, these sound more like nitpicks as long as #2 isn't an issue. 

> 1. Comment vs. implementation mismatch on retries. HexUploadPublisher.hpp and NRMAHexUploader.h describe reportId as "the persisted retry-counter key" / a "per-report retry limit," but NRMARetryTracker is explicitly in-memory only and resets every launch. There's no persisted counter. Either fix the comments to say retries are per-launch, or implement the persisted counter they imply. Related: an online report hitting repeated 5xx is dropped after 2 retries within a launch (finishWith:YES) — please confirm that transient-error drop is intended vs. keeping it for next launch.
>2. Duplicate-upload safety depends on the collector. "Keep until confirmed" means a termination after the server accepts but before the completion fires will re-upload next launch. That's only de-duped if the collector honors X-NewRelic-HexReport-Id. Please confirm collector-side dedup is actually live.
>3. HexStore::clear() prefix match is too loose. it->rfind(storePath, 0) == 0 will also match sibling directories sharing the prefix (e.g. /hex matches /hex2/...). Append a trailing separator (or match the parent dir exactly) before erasing claims.

cdillard-NewRelic (approved):
👍 I reran this and spammed hex and terminations and it appears to be working quite well. I didn't notice any issues with data correctness or duplicates.

diegomtz5 (approved):

PR #796, opened 2026-07-02
NR-581396 task-keyed payloadByTaskId, per-payload attempts counter, handledErroredTask:payload:.

mbruin-NR:
Handled-exception ("hex") reports were being uploaded to the collector more than once. Reproducible by generating several handled exceptions, backgrounding → foregrounding the app, then force-quitting: on the next launch the same reports were re-sent, producing duplicates in the UI.

## Fix

`Agent/HandledException/NRMAHexUploader.m`:
- Replace `payloadByRequest` (keyed by `NSURLRequest`) with `payloadByTaskId`, keyed by `@(task.taskIdentifier)`, which is unique within an `NSURLSession`. Each upload now gets its own slot; completions always resolve to their own payload.
- Move the retry counter onto the `NRMAHexPayload` (`attempts`) and drop the request-keyed `NRMARetryTracker` usage (it was broken the same way — retry counts were shared across unrelated reports). The payload survives across a retry and is re-registered under the rebuilt task's new identifier.
- Rename `handledErroredRequest:` → `handledErroredTask:payload:`.

Retry semantics (`kNRMARetryLimit = 2`), offline deferral, oversized-payload handling, and the concurrency cap are all unchanged. No public API changes.

`NRMARetryTracker.{h,m}` are now unused by production code but retained (still covered directly by `TestRetryTracker.m`); removing them can be a follow-up.

### DIFF
--- a/Agent/HandledException/NRMAHexUploader.m
+++ b/Agent/HandledException/NRMAHexUploader.m
@@ -7,7 +7,6 @@
 //
 
 #import "NRMAHexUploader.h"
-#import "NRMARetryTracker.h"
 #import "NRLogger.h"
 #include <libkern/OSAtomic.h>
 #import "NRMASupportMetricHelper.h"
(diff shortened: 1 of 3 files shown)


cdillard-NewRelic (approved):
👍 I tested this with backgrounding and termination scenarios and am not seeing duplicates. This appears to be fixed once changed to use the taskID as key.

PR #811, opened 2026-07-13
NR-589704 prevent old session replay frames from uploading with new session

mbruin-NR:
Added a clearData to session replay start to prevent old frames from being added to new

### DIFF
--- a/Agent/JSError/JSErrorController.swift
+++ b/Agent/JSError/JSErrorController.swift
@@ -179,11 +179,6 @@ public class JSErrorController: NSObject {
         errorQueueLock.lock()
         errorQueue.add(errorData)
         errorQueueLock.unlock()
-        
-        // Notify Session Replay of the error
-        if let agent = NewRelicAgentInternal.sharedInstance() {
-            agent.sessionReplay(onError: nil)
-        }
 
         // Persist to disk
         persistError(errorData)
--- a/Agent/Public/NewRelic.m
+++ b/Agent/Public/NewRelic.m
@@ -176,10 +176,10 @@
     if([NewRelicAgentInternal sharedInstance].isShutdown) {
         return;
     }
-
-    [[NewRelicAgentInternal sharedInstance].handledExceptionsController recordHandledException:exception];
     
     [[NewRelicAgentInternal sharedInstance] sessionReplayOnError:nil];
+
+    [[NewRelicAgentInternal sharedInstance].handledExceptionsController recordHandledException:exception];
 }
 
 + (void) recordHandledException:(NSException*)exception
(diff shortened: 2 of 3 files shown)


cdillard-NewRelic (changes requested):
Every other mutation of rawFrames/fullSnapshotFrameIndices in this class goes through the serial frameQueue:

This race is technically latent already (the existing external callers at NewRelicAgentInternal.m:725 and SessionReplayManager.clearAllData() use sessionReplayQueue, not frameQueue), but this change adds a new, hotter exposure. Route the in-memory clear through frameQueue, matching getAndClearFrames:

```
public func clearAllData() {
    frameQueue.async { [weak self] in
        guard let self = self else { return }
        self.rawFrames.removeAll()
        self.fullSnapshotFrameIndices.removeAll()
...
```

cdillard-NewRelic:
And from Claude: 


> 2. clearAllData() wipes the entire frames directory — can it eat previous-session recovery?
>clearAllData() → clearFramesDirectory() (:801) does removeItem(at: framesDirectory) — the whole kNRMA_SessionReplayFrames_folder, including other sessions' subdirs and their _upload_url.txt files. That's the persisted data checkForPreviousSessionFiles() uploads for crashed/previous sessions.

>At cold launch both run on the concurrent sessionReplayQueue with no ordering guarantee:
> - checkForPreviousSessionFiles is dispatched at init (NewRelicAgentInternal.m:387)
> - sessionReplayStart → start() → clearAllData() is triggered later, after harvest connect (NRMAHarvester.mm:984)

>In practice the recovery scan probably wins (it runs seconds earlier), which is why this likely passed testing — but "probably wins by timing" on a concurrent queue is exactly the kind of thing that regresses silently. Worth confirming: does clearing the whole directory at start() ever destroy a prior error-mode/crashed session's buffered frames before they're uploaded? If the only goal is the ticket's in-memory leak, clearing rawFrames (in-memory) is sufficient and the disk wipe is unnecessary risk here.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-589890

cdillard-NewRelic:
> 🟡 3. There's already a dedicated mechanism for this — is the blanket clear the right layer?

> The code has an error-mode marker (kNRMA_SessionReplayErrorMode_marker, :532) + NRMAExceptionHandlerManager specifically built to detect and clear a stale error-mode buffer on the next launch when no crash occurred. And transistionToRecordingMode's .error branch literally has the comment "When entering error mode, clear any existing frames" (:590) — but no clearing actually happens there; it only sets lastFullSnapshotTime. So the intended clear point was arguably the mode transition (or session end), and it was never implemented.

> Suggestion: consider whether the minimal, most correct fix is clearing the in-memory buffer on session end (endSession/stop) or on the error→something transition, rather than defensively at start(). Clearing at start() works, but it's clearing "the previous session's mess" at the start of the new one, which is why it collides with #2.

cdillard-NewRelic (approved):
Thanks for entertaining the claude comments I found interesting: I think this is much better now and handles the threading with a bit more sanity. Thanks for commenting out / removing the unused excessive logs. I ran extensive testing with NRTestApp-2 in error mode and things appear to be working well and I no longer see the original old session replay frame issue.

diegomtz5 (approved):

PR #813, opened 2026-07-15
NR-590263 rename MobileJSError event type to MobileError

diegomtz5:
React Native JS errors from iOS should report as MobileError instead of MobileJSError, ahead of the RN symbolicated-errors GA.


mbruin-NR (approved):

cdillard-NewRelic (approved):

mbruin-NR:
we are not changing from MobileJSError to MobileError

PR #820, opened 2026-07-22
NR-586741 setUserId needs to stay synchronous

mbruin-NR:
I added a breadcrumb to after setUserId and it was being added to the wrong session. That is why it needs to stay synchronous.

### DIFF
--- a/Agent/General/NewRelicAgentInternal.m
+++ b/Agent/General/NewRelicAgentInternal.m
@@ -874,6 +874,24 @@ static const NSString *kNRMA_APPLICATION_WILL_TERMINATE =
     [self onSessionStart];
 }
 
+- (void) startNewSessionForUserId:(NSString* _Nullable)userId {
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        @synchronized(kNRMA_BGFG_MUTEX) {
+            [self.analyticsController newSession];
+            [self sessionReplayEndSession];
+            [NewRelicAgentInternal harvestNow];
+            [self sessionStartInitialization];
+            if (userId) {
+                [self.analyticsController setSessionAttribute:kNRMA_Attrib_userId
+                                                        value:userId
+                                                   persistent:YES];
+            } else {
+                [self.analyticsController removeSessionAttributeNamed:kNRMA_Attrib_userId];
+            }
+        }
+    });
+}
+
 - (void) handle4HourSessionRestart {
     NRLOG_AGENT_DEBUG(@"Executing 4-hour automatic session restart");
 
--- a/Agent/General/NewRelicAgentInternal.m
+++ b/Agent/General/NewRelicAgentInternal.m
@@ -875,21 +875,19 @@ static const NSString *kNRMA_APPLICATION_WILL_TERMINATE =
 }
 
 - (void) startNewSessionForUserId:(NSString* _Nullable)userId {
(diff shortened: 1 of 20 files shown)


Agent/General/NewRelicAgentInternal.m
cdillard-NewRelic, line 319:
```suggestion
- (void) initialize {
    // Warm the identifierForVendor cache on this (startup) thread so that any later
    // session restart — which runs off the main thread (see startNewSessionForUserId:) —
    // is guaranteed a cache hit and never issues the identifierForVendor XPC call during
    // a sensitive lifecycle transition. NRMAUDIDManager caches the value after first read.
    (void)[NRMAUDIDManager deviceIdentifier];
```

Agent/General/NewRelicAgentInternal.m
```diff
@@ -875,21 +875,19 @@ - (void) sessionStartInitialization {
...
+        [self.analyticsController newSession];
+        [self sessionReplayEndSession];
+        [NewRelicAgentInternal harvestNow];
+        [self sessionStartInitialization];
+        if (userId) {
+            [self.analyticsController setSessionAttribute:kNRMA_Attrib_userId
+                                                    value:userId
+                                               persistent:YES];
+        } else {
+            [self.analyticsController removeSessionAttributeNamed:kNRMA_Attrib_userId];
```
cdillard-NewRelic, line 888:
```suggestion
    // The caller (setUserId:) has already assigned agent.userId synchronously, so
    // getUserId() reflects the new value immediately. The restart itself is offloaded to
    // a dedicated serial queue so the caller — frequently the main thread during a
    // lifecycle transition such as sceneWillResignActive — is never blocked by the
    // harvest + measurements re-initialization below. A serial queue also guarantees
    // that concurrent setUserId: calls are restarted in order rather than interleaving.
    //
    // The identifierForVendor cache is warmed on a safe thread at startup (see
    // -initialize), so nothing on this queue issues the identifierForVendor XPC call.
    static dispatch_queue_t sessionRestartQueue;
    static dispatch_once_t onceToken;
    dispatch_once(&onceToken, ^{
        sessionRestartQueue = dispatch_queue_create("com.newrelic.session.restart", DISPATCH_QUEUE_SERIAL);
    });

    dispatch_async(sessionRestartQueue, ^{
        @synchronized(kNRMA_BGFG_MUTEX) {
            // End the current session and flush its data under the previous userId BEFORE
            // tearing down the harvester. Execute the current harvester directly so the
            // final harvest is strictly ordered ahead of the teardown in
            // sessionStartInitialization; harvestNow only enqueues asynchronous work on
            // the harvest queue and cannot guarantee that ordering.
            [self.analyticsController newSession];
            [self sessionReplayEndSession];

            NRMAHarvester* harvester = [[NRMAHarvestController harvestController] harvester];
            if (harvester) {
                [harvester execute];
            }

            [self sessionStartInitialization];
            if (userId) {
                [self.analyticsController setSessionAttribute:kNRMA_Attrib_userId
                                                        value:userId
                                                   persistent:YES];
            } else {
                [self.analyticsController removeSessionAttributeNamed:kNRMA_Attrib_userId];
            }
        }
    });
```

mbruin-NR: I think this would return the issue that this change addresses.

cdillard-NewRelic (approved):

cdillard-NewRelic:
Okay changes declined.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-597235

PR #822, opened 2026-07-23
NR-597379 changing session replay from Timer to a DispatchSourceTimer

mbruin-NR:
https://new-relic.atlassian.net/browse/NR-597379
Replaced Timer + RunLoop.current.run() with DispatchSourceTimer. A DispatchSourceTimer fires directly onto the dispatch queue, requires no run loop, and releases the thread immediately after each tick. Key properties of this approach:
- DispatchSource serializes its own event handler even on a concurrent queue, so no concurrent timer firings
- cancel() + nil assignment cleanly stops it with no lingering state
- isRunning() now just checks sessionReplayTimer != nil

### DIFF
--- a/.github/workflows/uploadApp.yml
+++ b/.github/workflows/uploadApp.yml
@@ -15,13 +15,16 @@ on:
  # enables option for workflow to be manually executed in Github UI
  workflow_dispatch:
 
-#  push:
-#    branches:
-#      - main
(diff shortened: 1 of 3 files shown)


diegomtz5 (dismissed):

cdillard-NewRelic (dismissed):
Sorry I accidentally pushed to your branch. Approved.

cdillard-NewRelic (approved):

PR #828, opened 2026-07-28
NR-598682 sessionAttributeJSONString now takes a snapshot of dicts

mbruin-NR:
Fix in NRMASAM.mm: sessionAttributeJSONString now takes a snapshot of both dicts under the @synchronized blocks, releases the locks, then runs JSON serialization on the snapshot outside the lock. The profiler can still fire during serialization — it just no longer holds privateAttributeDict when it does, which breaks the cycle.

The random walk made this easy to reproduce because it combines rapid UI layout changes (triggering the TraceController path) with session replay activity (triggering the NRMASAM path) at the same time.

### DIFF
--- a/Agent/Analytics/NRMASAM.mm
+++ b/Agent/Analytics/NRMASAM.mm
@@ -224,29 +224,32 @@
 
 // Includes Public and Private Attributes
 - (NSString*) sessionAttributeJSONString {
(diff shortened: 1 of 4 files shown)


cdillard-NewRelic (approved):
Ran the walk tests, looks good and good test solution.

diegomtz5 (approved):

PR #831, opened 2026-07-28
NR-546764 clamp SwiftUI text width-compensation offset to its content…

diegomtz5:
… bounds

The space-compensation offset added to a captured SwiftUI Text's wireframe width scaled with the string's space count and was never bounded, letting long NavigationLink labels grow past their row's real content width and overlap the trailing disclosure chevron in session replay. Clamp it to the ambient SwiftUI content frame so it can widen a label only as far as its container actually allows.

Adds a NavigationLink Labels test case to NRTestApp to reproduce the issue.


diegomtz5:
this replay shows the fix https://staging.onenr.io/08wp2q24xjO

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-598857

mbruin-NR (approved):

cdillard-NewRelic (approved):

PR #837, opened 2026-08-03
Convert NewRelic public API to Swift POC

diegomtz5:
Pilot conversion of the public, customer-facing `NewRelic` API (`Agent/Public/NewRelic.h/.m`) from Objective-C to Swift — the first subsystem in a broader effort to migrate the agent to Swift.

- `NewRelic.m` deleted; `NewRelic.swift` is the new implementation
- `NewRelic.h` collapsed to a thin stub (re-exports the sibling declaration headers, the generated Swift interface, and 5 preserved ObjC helper macros) so `#import <NewRelic/NewRelic.h>` keeps compiling unchanged
- New `NRExceptionCatcher` (.h/.m) — a small ObjC bridge for the two `@try`/`@catch` call sites, since Swift cannot catch `NSException`
- All 73 selectors preserved exactly, including 8 undocumented ones (`setPlatformVersion:`, `startTracingMethodNamed:objectNamed:timer:category:`, etc.) that the six hybrid SDKs depend on but were never in the public header

Behaviour is intended to be identical for every existing caller in every language. This is not a redesign of the API surface.

## Please review these

- **`endTracingMethod(withTimer:)`** derives its `objc_setAssociatedObject` key via `Unmanaged.passUnretained(kNRTraceAssociatedKey as NSString).toOpaque()`, which must stay pointer-identical to the `(__bridge const void *)` casts in `NRMATraceController.m` / `NRMACustomTrace.m`. A mismatch fails silently. Now covered by `testSwiftEndTracingMethodWithTimerClearsObjCAssociatedKey`, verified by mutation.
- **`logErrorObject` / `recordError`** have `@objc` methods on concrete `NSError` plus `@nonobjc` `Error`-typed overloads. Intentional, so Swift-native error types keep bridging as they did when this file was ObjC.
- **Nullability.** Parameters the original header marked `_Null_unspecified` are implicitly-unwrapped optionals here, matching what Swift's importer already gave callers. Five methods that dereferenced those parameters without checking are now guarded — see below.

## Fixes found while verifying

- **Five nil-dereference crashes.** `endTracingMethodWithTimer:` and both `noticeNetworkRequest`/`noticeNetworkFailure` pairs dereferenced an implicitly-unwrapped parameter, so a nil argument trapped where the ObjC absorbed it (`[timer stopTimer]` and `[NSMutableURLRequest requestWithURL:]` both tolerate nil). Reachable from Unity, which pairs `createAndStartTimer` with `endTracingMethodWithTimer`. A Swift trap is not an `NSException`, so `@try` cannot catch it and no crash report is written.
- **No KVC left.** `crossProcessId()`, `currentSessionId()` and `recordJavascriptError()` previously reached agent internals through `value(forKey:)`. #895 added the ObjC changes that made them reachable directly, so all four string-keyed lookups are gone and every call is compiler-checked.
- **`testCrossProcessId` / `testCurrentSessionId`** compared `NSString` pointers via `XCTAssertEqual`, which only held while the values came back through KVC untouched. Now assert on value.
- All 33 hand-written `NRLogger.log(...)` calls replaced with the `NRLOG_*` helpers (#895, #898).

## Testing

- CI green on all three platform jobs; tvOS and watchOS suites run for the first time
- Full suite: 806 tests, only the 7 known network-dependent failures (`NRMASessionExclusivityWithDelegateTests` ×5, `NRMAWKNavigationDelegateBaseTest` ×2)
- Release XCFramework builds; `.swiftinterface` emitted for both slices with no `NewRelicPrivate` leakage; Swift and ObjC consumer smoke tests compile against it
- Crash reporting end-to-end: `crashNow` → PLCrashReporter → 200 from the staging crash collector
- Hybrid surface: the union of all six SDKs' public API driven against the built framework — 45 React Native and 19 Flutter call shapes, including the runtime-dispatched `setPlatformVersion:` verified on the wire as `platform = Flutter; platformVersion = "1.2.9"`
- 73/73 selectors resolve on the shipped binary via the ObjC runtime


Agent/Public/NewRelic.swift
```diff
@@ -0,0 +1,827 @@
...
+import Foundation
+@_implementationOnly import NewRelicPrivate
+
+@objcMembers
+public class NewRelic: NSObject {
+
+    // MARK: - Helpers for trying out New Relic features
+
+    @objc(crashNow:)
+    public static func crashNow(_ message: String?) {
```
cdillard-NewRelic: Can we move this to a newRelic+Helpers.swift extension so the header doesn't start with crash? Nice changes.

diegomtz5: I agree I will move it over.

Agent/Public/NewRelic.swift
```diff
@@ -316,25 +316,44 @@ public class NewRelic: NSObject {
...
+        // The original header annotated selector, object and timer `_Null_unspecified`, which
+        // the Swift importer surfaced to callers as Selector!/Any!/NRTimer!. Declaring the
+        // first two non-optional narrowed that and broke existing Swift callers (a value read
+        // out of an IUO-returning call infers as a plain Optional, so it no longer binds), so
+        // the implicitly-unwrapped optionals are restored here.
+        //
+        // Guard rather than implicitly unwrap: the ObjC implementation tolerated nil arguments,
+        // so trapping on them would be a behaviour change in the opposite direction.
+        guard let selector = selector, let object = object else {
+            NRLogger.log(NRLogLevelVerbose.rawValue, inFile: #fileID, atLine: UInt32(#line), inMethod: #function, withMessage: "\(#function) called with a nil selector or object; ignoring.", withAgentLogsOn: true)
```
cdillard-NewRelic, line 329: I see this
```
NRLogger.log(NRLogLevelVerbose.rawValue, inFile: #fileID, atLine: UInt32(#line), inMethod: #function, withMessage: "\(#function) called with a nil selector or object; ignoring.", withAgentLogsOn: true)
```
In a few spots, Don't we have a swift helper for logging with nicer syntax? if not no worries. Does NRLOG_VERBOSE work for this purposes?

diegomtz5: Yes there is, I'll clean those up. also working on cleaning these calls `guard let jsErrorController = agent.value(forKey: "jsErrorController") as? JSErrorController else {`

cdillard-NewRelic: I guess it should be NRLOG_AGENT_VERBOSE, that might need to be added. Im working to add better details about what is console, remote, agent logs.

diegomtz5: Yeah I am starting another PR with all the changes needed for this one to work just need to add one more thing before opening it, here is the [branch](https://github.com/newrelic/newrelic-ios-agent/tree/refactor/swift-consumable-agent-internals )

cdillard-NewRelic:
Super cool to see this reduce the number of lines of code by 860! Great work.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-619465

PR #850, opened 2026-08-07
NR-602919 Fix session replay events sorting ahead of Meta/FullSnapshot

diegomtz5:
**Bugs found and fixed**

1. Meta event displaced by an early touch — the rrweb player requires Meta to be the first event in the stream to initialize the viewport. A touch sorted ahead of it produced "First event didn't include meta" in the player.
2. FullSnapshot displaced by an early touch — FullSnapshot establishes the DOM node IDs that later touch/incremental events reference. A touch sorted ahead of it broke replay for the same reason.
3. Empty-frame harvest producing a touches-only payload — if a harvest boundary (60s timer, backgrounding, or a setUserId-triggered new session) fires before the first frame has finished its async capture, touches were being uploaded with no frame/meta at all to anchor them to.

Both the live-harvest path (SessionReplayManager.mergeAndSortReplayEvents) and the crash-recovery path (NRMASessionReplay.processFrameToFile) had the same displacement issue and were fixed the same way: anchor Meta (and FullSnapshot, when present) at the front of the merged event list, then sort everything else normally.

**Testing**

- Added regression tests (SessionReplayHarvestEndToEndTests.swift) covering touch-before-meta and touch-before-fullsnapshot through the real harvest path.
- Manually reproduced all three bugs on-device against the staging collector and confirmed each is fixed, including a deterministic repro for bug 3 (rapid back-to-back setUserId calls before the first frame captures).

Fixes NR-602919.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,3 +93,4 @@ Test Harness/NRTestApp/upload_dsym_results.log
 node_modules
 derived_path
 builds/nrtestapp-ios.zip
+.worktrees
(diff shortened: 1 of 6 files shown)


mbruin-NR (approved):

cdillard-NewRelic (approved):
I've validated your PR by using Charles and checking that each /blobs request starts with a type 4 (meta event). Great work. 👍

PR #854, opened 2026-08-10
NRMAExceptionMetaDataStore rewritten by Claude

cdillard-NewRelic:
Claude: 

> Seven files, all inside the crash-metadata blast radius. The core move: the 14 string fields are now fixed-capacity double-buffered slots published by one release store and read by one acquire load. Nothing is allocated or freed, so a double free is no longer expressible and a reader can never hold a dangling pointer. Each slot carries an untouched trailing guard NUL, so every read is bounded even mid-race. Also fixed: the temp-path overflow (plus a NULL-tempDir deref), NRMA_write returning the last chunk instead of the total (every caller compares against len, so a split write silently truncated the file), and the interaction history — an unsynchronized malloc'd linked list, read and freed from the handler and unbounded for the process lifetime — is now a bounded ring.
> 
> Net: the crash handler path is now entirely free of malloc, free, locks, and Objective-C. The one remaining malloc is in NRMA_createTempFileName, which the handler no longer calls.
> 
> Tests: 12/12 in NRMAExceptionMetaDataTest, including the pre-existing byte-exact format assertion (wire format unchanged). I corrected one test whose claim I couldn't substantiate — blocking descriptors don't short-write, so I could not build a non-flaky test that fails on the old NRMA_write; that one now honestly pins the contract instead of pretending to reproduce a split.


diegomtz5:
Nice work! I reviewed it against our fix, rebuilt the branch and ran NRMAExceptionMetaDataTest (12/12 pass), everything looks good.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-604012

diegomtz5 (approved):

mbruin-NR (approved):

PR #868, opened 2026-08-14
Capture server errors and config update

mbruin-NR:
Introduces two related capabilities to the capture test harness:

Error injection
A new "Inject" button in the captures toolbar arms the server to return a chosen HTTP error code (429, 5xx) for a specific endpoint and a configurable number of times. While an injection is active the button shows the armed code as a pill badge. The inject sheet supports one-shot, repeat-N, and unlimited modes, with a Disarm option to cancel early.

Connect config editing
The connect response is now modeled as a typed ConnectConfig struct (Codable, replaces the hardcoded [String: Any] dict). A new button in the captures toolbar opens a dedicated config sheet where session replay settings (enabled, sampling rates, mode, mask toggles, custom masking rules) and log settings (enabled, level, sampling rate) can be edited with standard form controls. Changes apply immediately to the live server. Saving automatically arms a single 409 on /data so the agent reconnects and picks up the new settings — this is the only path that reliably triggers a reconnect, since a 409 on /connect is silently ignored by the harvester's state machine.

Behavioral verification
verifyAll() now tracks the effective ConnectConfig at each capture's timestamp and adds a failing check to any upload that shouldn't have happened given that config — e.g. a /mobile/blobs upload after session_replay.enabled=false, or a /mobile/logs upload after logs.enabled=false. These checks appear inline on the offending captures in the detail view.


Test Harness/NRTestApp/NRTestApp/CaptureServer/ConnectConfig.swift
```diff
@@ -0,0 +1,116 @@
...
+
+    /// Returns a [String: Any] dict ready to hand to HttpServer as the connect response body.
+    /// Injects `at_capture` ([1, []]) which can't be round-tripped through Codable cleanly.
+    func toServerDict() -> [String: Any] {
+        let enc = JSONEncoder()
+        guard let data = try? enc.encode(self),
+              var dict = (try? JSONSerialization.jsonObject(with: data)) as? [String: Any] else {
+            return [:]
+        }
+        dict["at_capture"] = [1, []] as [Any]
```
cdillard-NewRelic: what is at_capture? as far as i can tell its gone from server connect responses now. 
```
{
	"data_token": [246132793, 339923787],
	"account_id": "10816994",
	"trusted_account_key": "1",
	"entity_guid": "MTA4MTY5OTR8TU9CSUxFfEFQUExJQ0FUSU9OfDI0NjEzMjc5Mw",
	"request_headers_map": {
		"NR-Session": "Bvq2rHpSm7taAKUN4hRC00sAAAEBLCQoAAAAHQECBA6rsDnIAAAAAWp_fcEDAANERVYAAAAJTlJUZXN0QXBw",
		"NR-AgentConfiguration": "AQAAAZ/yp61Y"
	},
	"configuration": {
		"logs": {
			"enabled": true,
			"level": "DEBUG",
			"sampling_rate": 100.0
		},
		"application_exit_info": {
			"enabled": true
		},
		"session_replay": {
			"enabled": true,
			"mode": "custom",
			"sampling_rate": 100.0,
			"error_sampling_rate": 100.0,
			"mask_application_text": false,
			"mask_user_input_text": false,
			"mask_all_user_touches": false,
			"mask_all_images": false,
			"custom_masking_rules": [{
				"type": "mask",
				"operator": "equals",
				"name": ["msr-swiftui-text-1"],
				"identifier": "accessibilityIdentifier"
			}, {
				"type": "unmask",
				"operator": "equals",
				"name": ["msr-swiftui-text-2"],
				"identifier": "accessibilityIdentifier"
			}]
		}
	}
}
```

mbruin-NR: https://github.com/newrelic/newrelic-ios-agent/blob/1e002caf2b18ed018e0c48ce8de625f9c5704b2d/Agent/Harvester/DataStore/NRMAHarvesterConfiguration.m#L95
I think it's suppose to hold the maxTotalTraceCount and the activityTraceConfigurations which holds the activityTraceNamePattern and totalTraceCount. But I guess it was removed at some point.

cdillard-NewRelic: yeah it was removed from /connect response and I noticed some of our code relied on it still. I will make a PR fixing it. I think we hsould remove it from PR and match wht existing /connect returns.

cdillard-NewRelic (approved):
👍

diegomtz5 (approved):

PR #870, opened 2026-08-17
NR-323614 added a check to not allow arrays or dictionaries to the new event system

mbruin-NR:
The legacy C++ event system enforced this restriction structurally: setSessionAttribute:value: and the event:withAttributes: bridge method used an explicit if/else if type-dispatch that only had branches for NSNumber, NSString, and NRMABool. Arrays and dictionaries fell to the final else, logged an error, and returned NO — the restriction came for free from the bridge's inability to translate those types to C++ primitives or const char*.

The new event system delegates validation to NRMAAttributeValidator, which previously caught arrays and dictionaries only incidentally — they fell through the NSString check and hit the generic else if (![value isKindOfClass:[NSNumber class]] && ...) catch-all. This worked correctly but gave a misleading error message ("value must be an NSString, NSNumber, or BOOL") with no explicit mention of what type was actually passed.

This PR adds an explicit early-exit check for NSArray and NSDictionary in NRMAAttributeValidator.valueValidator: with a targeted error message, and adds unit tests in NRMASAMTest (session attributes) and TestCustomEvents (event attributes) that use the production NRMAAttributeValidator to confirm both types are rejected.

### DIFF
--- a/Agent/Analytics/AttributeValidator/NRMAAttributeValidator.m
+++ b/Agent/Analytics/AttributeValidator/NRMAAttributeValidator.m
@@ -55,6 +55,11 @@
         return false;
     }
 
+    if ([value isKindOfClass:[NSArray class]] || [value isKindOfClass:[NSDictionary class]]) {
+        NRLOG_AGENT_ERROR(@"invalid attribute: value cannot be an NSArray or NSDictionary");
+        return false;
+    }
+
     if ([value isKindOfClass:[NSString class]]) {
         if ([(NSString*)value length] == 0) {
             NRLOG_AGENT_ERROR(@"invalid attribute: value length = 0");
--- a/Test Harness/NRTestApp/NRTestApp/ViewModels/UtilViewModel.swift
+++ b/Test Harness/NRTestApp/NRTestApp/ViewModels/UtilViewModel.swift
@@ -59,6 +59,10 @@ class UtilViewModel {
         options.append(UtilOption(title: "End Interaction Trace", handler: { [self] in stopInteractionTrace()}))
         options.append(UtilOption(title: "Notice Network Request", handler: { [self] in noticeNWRequest()}))
         options.append(UtilOption(title: "Notice Network Failure", handler: { [self] in noticeFailedNWRequest()}))
+        
+        // NR-323614 — NSArray/NSDictionary values must be rejected as attribute values.
+        options.append(UtilOption(title: "Try Collection Session Attributes", handler: { [self] in setCollectionSessionAttributes()}))
+        options.append(UtilOption(title: "Try Collection Event Attributes", handler: { [self] in setCollectionEventAttributes()}))
 
         options.append(UtilOption(title: "Test System Logs", handler: { [self] in testSystemLogs()}))
         options.append(UtilOption(title: "Notice Network Request w headers/params", handler: { [self] in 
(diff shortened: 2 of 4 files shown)


cdillard-NewRelic (dismissed):
Ran tests and NRTestApp locally - passed.

Looks good - could consider addition to UtilViewModel 

```
        // NR-323614 — NSArray/NSDictionary values must be rejected as attribute values.
        options.append(UtilOption(title: "Set Collection User Attributes (NR-323614)", handler: { [self] in setCollectionUserAttributes()}))
        options.append(UtilOption(title: "Set Collection Session Attributes (NR-323614)", handler: { [self] in setCollectionSessionAttributes()}))
        options.append(UtilOption(title: "Set All Collection Attributes (NR-323614)", handler: { [self] in setAllCollectionAttributes()}))
```

```
    // MARK: - NR-323614: collection values are invalid attribute values

    /// The four collection types NRMAAttributeValidator must reject, plus the Swift
    /// collection literals that bridge to NSArray/NSDictionary at the ObjC boundary.
    private var collectionAttributeValues: [(label: String, value: Any)] {
        let nsArray: NSArray = NSArray(array: ["one", "two"])
        let nsMutableArray = NSMutableArray(array: ["one", "two"])
        let nsDictionary: NSDictionary = NSDictionary(dictionary: ["key": "value"])
        let nsMutableDictionary = NSMutableDictionary(dictionary: ["key": "value"])

        return [
            ("NSArray", nsArray),
            ("NSMutableArray", nsMutableArray),
            ("NSDictionary", nsDictionary),
            ("NSMutableDictionary", nsMutableDictionary),
            ("SwiftArray_bridgedToNSArray", ["one", "two"]),
            ("SwiftDictionary_bridgedToNSDictionary", ["key": "value"]),
        ]
    }

    /// Tries every collection type as a user (global, persistent) attribute.
    /// Exercises `NewRelic.setAttribute` -> `NRMAAnalytics setSessionAttribute:value:persistent:YES`
    /// -> `NRMASAM setAttribute:value:` -> `NRMAAttributeValidator valueValidator:`.
    /// Every call is expected to return false and log
    /// "invalid attribute: value cannot be an NSArray or NSDictionary".
    func setCollectionUserAttributes() {
        NewRelic.enableFeatures([.NRFeatureFlag_NewEventSystem])

        var results = [String]()
        for sample in collectionAttributeValues {
            let name = "collectionUserAttr_\(sample.label)"
            let didSet = NewRelic.setAttribute(name, value: sample.value)
            results.append(verdict(name: name, accepted: didSet, shouldAccept: false))
        }

        // Control case: a valid value must still be accepted after the rejected ones.
        let didSetValid = NewRelic.setAttribute("collectionUserAttr_validString", value: "good attribute")
        results.append(verdict(name: "collectionUserAttr_validString", accepted: didSetValid, shouldAccept: true))

        report("user attributes via NewRelic.setAttribute", results)
    }

    /// Tries every collection type as a session attribute carried on session events.
    /// Exercises `NRMACustomEvent addAttribute:value:` (custom events and breadcrumbs) with the
    /// same validator. The event itself is expected to record while the collection attribute is
    /// dropped, so one bad attribute no longer poisons the whole harvest batch.
    func setCollectionSessionAttributes() {
        NewRelic.enableFeatures([.NRFeatureFlag_NewEventSystem])

        var results = [String]()
        for sample in collectionAttributeValues {
            let eventAttributes: [String: Any] = [
                "collectionSessionAttr": sample.value,
                "validSessionAttr": "good attribute"
            ]

            let didRecordEvent = NewRelic.recordCustomEvent("CollectionAttributeTest",
                                                            attributes: eventAttributes)
            let didRecordBreadcrumb = NewRelic.recordBreadcrumb("CollectionAttributeTest",
                                                                attributes: eventAttributes)
            results.append("\(sample.label): recordCustomEvent -> \(didRecordEvent ? "recorded" : "dropped"), recordBreadcrumb -> \(didRecordBreadcrumb ? "recorded" : "dropped") (collection attribute itself must be absent from the payload)")

            // Errors take the same event attribute path; it returns Void, so just make the call.
            do {
                try errorMethod()
            } catch {
                NewRelic.recordError(error, attributes: eventAttributes)
            }
        }

        report("session attributes via events/breadcrumbs/errors", results)
    }

    /// Runs both collection attribute checks back to back.
    func setAllCollectionAttributes() {
        setCollectionUserAttributes()
        setCollectionSessionAttributes()
    }

    private func verdict(name: String, accepted: Bool, shouldAccept: Bool) -> String {
        let outcome = accepted ? "accepted" : "rejected"
        let status = accepted == shouldAccept ? "PASS" : "FAIL"
        return "\(name): \(outcome) (expected \(shouldAccept ? "accepted" : "rejected")) [\(status)]"
    }

    private func report(_ title: String, _ results: [String]) {
        print("NR-323614 collection attribute test — \(title)")
        for result in results {
            print("  NR-323614  \(result)")
        }
        attributes = results.joined(separator: "\n")
    }

```

diegomtz5 (approved):

cdillard-NewRelic (approved):

PR #875, opened 2026-08-26
NR-610664 Trying to remove failure points from getting sessionAttributes into a JSError

mbruin-NR:
It really doesn't seem like the JSErrorController can be initialized without the analyticsController being initialized first. 
My best guess is a failure to encode and or decode the sessionAttributeJSONString. I attempted to fix this by adding a new direct dictionary path that avoids the JSON transition.

### DIFF
--- a/Agent/Analytics/NRMAAnalytics+cppInterface.h
+++ b/Agent/Analytics/NRMAAnalytics+cppInterface.h
@@ -11,6 +11,7 @@
 
 @interface NRMAAnalytics (cppInterface)
 - (std::shared_ptr<NewRelic::AnalyticsController>&) analyticsController;
+- (NSDictionary*) sessionAttributeDictionaryFromCpp;
 
 - (BOOL) recordLegacyEventResult:(NewRelic::EventAddResult)result;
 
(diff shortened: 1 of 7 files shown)


cdillard-NewRelic (approved):
Looks good to me , claude only flagged one nitpik about what happens if NewEventSystem flag is toggled after agent already started. 

```
Fix for cause A: dispatch on which store exists, not on the flag

The bug is that initWithSessionStartTimeMS: picks one store exclusively (_sessionAttributeManager in the new-event-system branch, _analyticsController in the else), but every accessor re-reads [NRMAFlags shouldEnableNewEventSystem] at call time. NewRelic.enableFeatures(...) is public and callable at any moment, so the two can disagree — and when they do:

- flag flipped on after start → _sessionAttributeManager is nil → [nil sessionAttributeDictionary] → nil → every JS error ships with zero attributes. This is your reported symptom.
- flag flipped off after start → _analyticsController is a default-constructed null shared_ptr → ->getSessionAttributes() is a null deref, SIGSEGV, uncatchable by the try/catch.

Fix both at once by asking what actually exists:

- (NSDictionary*) sessionAttributeDictionary {
    if (_sessionAttributeManager) {
        return [_sessionAttributeManager sessionAttributeDictionary];
    }
    if (_analyticsController) {
        return [self sessionAttributeDictionaryFromCpp];
    }
    NRLOG_AGENT_ERROR(@"sessionAttributeDictionary: no attribute store initialized");
    return @{};
}
```

Ill leave it to your discretion to handle now or we can make a ticket to handle.

PR #883, opened 2026-09-03
Harvester/NRMAHarvesterConfiguration LifeTime Fix

cdillard-NewRelic:
Root cause

NRMAHarvesterConfiguration declared eight object-typed properties as assign. Under ARC (both files compile with ARC — neither carries -fno-objc-arc) that means unsafe_unretained. initWithDictionary: pointed them straight into the autoreleased dictionary it was parsing — either the NRMAJSON graph from configureFromCollector: or the [NSUserDefaults objectForKey:] graph from fetchHarvestConfiguration. The configuration object itself is held in a strong ivar that long outlives that pool, so every later read was a use-after-free, surfacing as SIGSEGV in objc_msgSend.

The prior fix (#742) couldn't work: its [rawHeaders isKindOfClass:] guard is an objc_msgSend on the dangling pointer, and @catch cannot catch a SIGSEGV. That's why the crash survived 7.7.2 → 7.7.6, drifting from :549 to :678.

Changes

- NRMAHarvesterConfiguration.h — the eight properties assign → copy (encoding_key, request_header_map, log_reporting_level, session_replay_mode, and SessionReplayCustomMaskingRule's identifier/name/operatorName/type). session_replay_mode is now also typed as its real SessionReplayMaskingMode, which the Swift consumers in ViewDetails.swift require once ownership changes.
- NRMAHarvester.mm — removed the dead request_header_map workaround in configureHarvester:; corrected the misleading comment on the transitionToConnected: guard (I kept the guard — it provides real nil-safety when configureFromCollector: returns nil).
- NRMAHarvesterConfiguration.m — isEqual: compared application_token and request_header_map by raw pointer. application_token is the first check, so it short-circuited everything after it and left config-change detection permanently reporting "changed".
- NRMADeviceInformation.h — platformVersion was the identical assign NSString* defect.
- New NRMAHarvesterConfigurationLifetimeTests.m, registered in all three test targets.

Verification (zombies + MallocScribble via the .xctestrun, not the shared scheme)

Before, each test crashed the xctest process — note the PID changing as the runner relaunched:

testCollectorConfigurationOutlivesItsParseDictionary
  *** -[__NSDictionaryI retain]: message sent to deallocated instance 0x10656f2c0
testCustomMaskingRulesOutliveTheirParseDictionary
  *** -[__NSSingleObjectArrayI retain]: message sent to deallocated instance 0x101afcd70
testStoredConfigurationOutlivesItsUserDefaultsDictionary
  *** -[CFString retain]: message sent to deallocated instance 0x10bd58840
** TEST EXECUTE FAILED **

After: 74 tests, 0 failures — my 5 plus the existing Harvester, HarvestController, HarvesterState, HarvesterConnection, Connection, and SessionReplay suites.

### DIFF
--- a/Agent.xcodeproj/project.pbxproj
+++ b/Agent.xcodeproj/project.pbxproj
@@ -52,6 +52,7 @@
 		0209AC3124E7073800E45C90 /* NRMADeviceInformationTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 0209AC2D24E7073800E45C90 /* NRMADeviceInformationTests.m */; };
 		0209AC3224E7073800E45C90 /* NRMAHarvestableAnalyticsTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 0209AC2E24E7073800E45C90 /* NRMAHarvestableAnalyticsTest.m */; };
 		0209AC3424E7074100E45C90 /* NRMAConnectionTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 0209AC3324E7074100E45C90 /* NRMAConnectionTest.m */; };
+		CD11FE0002000000000000A2 /* NRMAHarvesterConfigurationLifetimeTests.m in Sources */ = {isa = PBXBuildFile; fileRef = CD11FE0001000000000000A1 /* NRMAHarvesterConfigurationLifetimeTests.m */; };
 		0209AC3D24E7075200E45C90 /* NRHarvestableVitalsTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 0209AC3624E7075100E45C90 /* NRHarvestableVitalsTest.m */; };
 		0209AC3E24E7075200E45C90 /* NRHarvesterTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 0209AC3724E7075100E45C90 /* NRHarvesterTest.m */; };
 		0209AC3F24E7075200E45C90 /* NRHarvesterStateTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 0209AC3824E7075200E45C90 /* NRHarvesterStateTest.m */; };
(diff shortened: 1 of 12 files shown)


ndesai-newrelic (commented):
potential crash on non-array/non-numeric collector fields still exists in config parsing.

- In `Agent/Harvester/DataStore/NRMAHarvesterConfiguration.m:103-105`, any present `kNRMA_AT_CAPTURE` value is passed directly to `initWithArray:`.
- In `Agent/Harvester/DataStore/NRMATraceConfigurations.m:29-31`, parsing immediately uses `array.count`; if collector sends `null`/`NSNull` or another non-`NSArray` type, this can become an unrecognized selector crash instead of a safe fallback.
- Similar risk exists for `KNRMA_AT_MIN_UTILIZATION` at `Agent/Harvester/DataStore/NRMAHarvesterConfiguration.m:110-111` where `doubleValue` is called without type checks.

ndesai-newrelic:
Review finding (medium): potential crash on non-array/non-numeric collector fields still exists in config parsing.

- In `Agent/Harvester/DataStore/NRMAHarvesterConfiguration.m:103-105`, any present `kNRMA_AT_CAPTURE` value is passed directly to `initWithArray:`.
- In `Agent/Harvester/DataStore/NRMATraceConfigurations.m:29-31`, parsing immediately uses `array.count`; if collector sends `null`/`NSNull` or another non-`NSArray` type, this can become an unrecognized selector crash instead of a safe fallback.
- Similar risk exists for `KNRMA_AT_MIN_UTILIZATION` at `Agent/Harvester/DataStore/NRMAHarvesterConfiguration.m:110-111` where `doubleValue` is called without type checks.

Suggested fix: add type guards (`isKindOfClass:[NSArray class]` / numeric checks) before calling parser/converters, and fallback to existing defaults when shape/type is unexpected. Also add tests for `at_capture: null` and non-numeric min-utilization values to lock behavior.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-614812

mbruin-NR (approved):

Issue #884, opened 2026-09-03
Crash while saving a handled-exception report on iOS 26.6 (7.7.4) — looks like #743 / #745 coming back


vishal-scapia:
# Crash while saving a handled-exception report on iOS 26.6 (7.7.4) — looks like #743 / #745 coming back

## Description

We got one fatal crash from a real user's phone. It happened inside the Hex code that saves a
handled-exception report. The agent version is `NewRelicAgent 7.7.4`.

That version already has both of your earlier fixes:

- #743 — added an empty check to `LibraryController::getAppImage()` and fixed a `call_once` race
- #745 — fixed file-handle leaks and a bad read in `HexStore.cxx`

Both shipped in 7.7.2, and we are on 7.7.4. So this looks like the same problem coming back in a
slightly different form.

The crash follows the exact path that #743 says it fixed:

```
recordError:attributes: → HexController::submit → HexReport::finalize
  → AgentData::serialize → HandledException::serialize → buildLibraries → getAppImage()
```

The trace contains `__cxa_call_terminate`. That means a C++ exception was thrown and nobody caught
it, so the app was killed. So something on this path is still throwing, either from a spot that the
`try/catch(...)` in #743 does not cover, or from a frame that has no guard at all.

**One thing to be upfront about: we could not reproduce this.** It is a single crash from a single
user, collected through Crashlytics. We are reporting it because this path has been fixed twice
already, and this looks like a case that got through. Tell us what data would help and we will get it.

## Steps to Reproduce

We do not have reliable steps. We have not been able to make it happen again in-house. Here is what
was true when it crashed:

1. Agent started with the config shown below. App was in the foreground, about 8 minutes into the session.
2. The phone was very low on memory — about 175 MiB free out of roughly 6.8 GiB in use.
3. A `recordError` call came in from Flutter over the method channel. That goes into
   `HexController::submit`, which then serializes the report.
4. The app crashed during that serialization.

If you want a guess at a synthetic test: call `recordError` in a tight loop on iOS 26.6 while the
device is nearly out of memory, so that the allocations inside `HexStore::store` and
`HandledException::serialize` start failing. **We have not tried this and cannot say it works.**

## Expected Behavior

The agent should not kill the app when it fails to save a report. If serialization or the disk write
fails, we would expect the report to be dropped and the app to keep running.

This is the same rule you applied on the Dart side in #228: the code that watches the app should
never be able to break the app.

[...]

PR #885, opened 2026-09-03
NR-614312 Stop a low-memory recordError from killing the host app

cdillard-NewRelic:
A std::bad_alloc raised while serializing a handled-exception report unwound out of -[NRMAHandledExceptions recordError:attributes:] with no handler above it, hit std::terminate, and killed the host app. Reported as GitHub issue #884, with the frames HexController::submit -> HexStore::store -> HexReport::finalize -> AgentData::serialize -> HandledException::serialize and __cxa_call_terminate.

One recordError call makes 112-304 heap allocations and any single failure was fatal, which is why it surfaced as one crash from one user on a device with ~175 MiB free of ~6.8 GiB in use.

Three changes:

- NRMAHandledExceptions.mm: the three public record* entry points are now thin guarded wrappers around nrma_*Unguarded: internals, logging and dropping the report on any C++ exception. The guard is here rather than inside HexController::submit because the throw can originate in createReport() and in the attribute plumbing, both of which run before submit is reached — guarding submit alone just relocates the crash to HexReport's constructor.

- HandledException.cxx: buildLibraries() and the thread loop no longer swallow exceptions raised while writing to the FlatBufferBuilder. Library::serialize and Thread::serialize open a table, so catching mid-write left the builder with nested_ == true and the next CreateVector tripped FLATBUFFERS_ASSERT(!nested) — an abort in debug, and a silently malformed buffer in release, where ENABLE_NS_ASSERTIONS=NO defines NDEBUG and compiles the assert out. Only the builder-free work stays guarded: copying the library snapshot and reading the app image UUID.

- HexStore::store(): serializes before opening the file, so a report that fails to serialize no longer leaves a 0-byte .fbad behind for readAll() to reap.

Adds Tests/OOM-Resilience and scripts/run_oom_resilience_tests.sh, which make the failure deterministic by replacing global operator new so 1-in-N allocations throw. Measured across this change:

  before  app killed by uncaught std::bad_alloc within ~2k recordError calls.
          Reports on disk included 0-byte files, reports truncated mid-library-
          list, and one buffer the flatbuffers verifier rejected outright.
  after   20,000 calls survived, 10,341 reports dropped and logged, no crash.
          50,000 calls survived in the native harness. Every stored report
          verifies, in one of two shapes: complete, or zero libraries when the
          library snapshot itself failed.

The suite ships a --expect-crash mode that asserts the app DOES die, so the test can be shown to still catch the original bug rather than passing vacuously.

Not addressed here: the agent's harvest and upload threads do unguarded C++ allocation too and die the same way under process-wide injection. That is a larger piece of work than this issue; Tests/OOM-RESILIENCE.md calls it out so it is not mistaken for covered ground.

### DIFF
--- a/Agent/HandledException/NRMAHandledExceptions.mm
+++ b/Agent/HandledException/NRMAHandledExceptions.mm
@@ -17,6 +17,7 @@
 #import "NRMAHarvestController.h"
 #import "NRMAAppToken.h"
 #include <execinfo.h>
+#include <exception>
 #import "NRMAFlags.h"
 #include <Analytics/AnalyticsController.hpp>
 #import "NRMABool.h"
@@ -33,6 +34,17 @@
 @end
 
 const NSString* kHexBackupStoreFolder = @"hexbkup/";
+
+// The unguarded bodies of the three public record* entry points. Every public
+// entry point wraps its counterpart here in the crash boundary below; nothing
+// else should call these directly.
+@interface NRMAHandledExceptions (CrashBoundary)
+- (void) nrma_recordErrorUnguarded:(NSError* _Nonnull)error
+                        attributes:(NSDictionary* _Nullable)attributes;
+- (void) nrma_recordHandledExceptionUnguarded:(NSException*)exception
+                                   attributes:(NSDictionary*)attributes;
+- (void) nrma_recordHandledExceptionWithStackTraceUnguarded:(NSDictionary*)exceptionDictionary;
+@end
 
 @implementation NRMAHandledExceptions {
     NewRelic::Hex::HexController* _controller;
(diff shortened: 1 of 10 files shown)


mbruin-NR (dismissed):

diegomtz5:
There may be a small issue with this, I'll just post Claude's findings:

Frame skipping is off by one on the recordError path

Splitting recordError:attributes: into a guarded wrapper + nrma_recordErrorUnguarded:attributes: moves backtrace() one frame deeper (now NRMAHandledExceptions.mm:285). The wrapper's [self nrma_recordErrorUnguarded:...] is a real objc_msgSend inside a try, so it can't be inlined or tail-called — the extra frame is always there. But createThreadVector:length: still discards a hardcoded 2 frames (NRMAHandledExceptions.mm:405), unchanged by this PR.

Net effect on every recordError report from upgraded apps: the reported leaf frame becomes -[NRMAHandledExceptions recordError:attributes:] (agent code) instead of the customer's calling function, and all customer frames shift one deeper. Since the trace fingerprints handled errors, existing error groups will re-fork on upgrade. recordHandledException: and recordHandledExceptionWithStackTrace: are unaffected — they use callStackReturnAddresses / a caller-supplied dict.

Verified with a test that captures the backtrace handed to createThreadVector:length: and resolves it via dladdr. Same test, same simulator — passes on develop, fails here:

[0] -[NRMAHandledExceptions nrma_recordErrorUnguarded:attributes:]   <- agent
[1] -[NRMAHandledExceptions recordError:attributes:]                 <- agent
[2] -[TestHandledExceptionController testRecordError...]             <- caller

Observed leading agent frames: 2, so the skip constant should be 3, but it is 2.

(develop shows 1 agent frame, so skip-2 is correct there. Both public entry points in NewRelic.m call the instance method directly, one frame above it.)

Suggested fix: capture backtrace() in the thick/frames into the unguarded body. Keeps theconstant at 2 and decouples it from call depth — bumping it to 3 also works but leaves the same latent coupling that broke here.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-617252

cdillard-NewRelic:
@diegomtz5 Thanks for the feedback and noticing that issue with the backtrace. I've added a fix so it gets rid of the New Relic frames in the error stack now.

diegomtz5 (approved):

PR #894, opened 2026-09-14
fix: send correctly-cased ReactNative platform value for JS error upl…

diegomtz5:
…oads

MobileErrorsUploader hardcoded platform=reactnative, diverging from the Android agent's ApplicationFramework.ReactNative and from the correct kNRMAPlatformString_ReactNative constant already threaded into JSErrorController. Pass that value through to the uploader instead of hardcoding a second, drifted copy.


ndesai-newrelic (approved):

cdillard-NewRelic (approved):
:+1:

PR #895, opened 2026-09-14
Make agent internals correctly consumable from Swift (removes the need for KVC lookups)

diegomtz5:
## Why

`+[NewRelic ...]`'s Swift port (#837) has to reach several agent internals that Swift cannot see or cannot name correctly, and it currently does so through `value(forKey:)` — string-keyed KVC. That is not a style problem. `value(forKey:)` with a key that no longer resolves raises `NSUnknownKeyException`, which **Swift cannot catch**, so renaming an ObjC property would crash the host app from inside a public API — and neither the compiler nor the test suite would notice. The React Native bridge's `recordJavascriptError` call site (`NRMModularAgent.mm:288`) has no `@try` guard, so that crash path is real today.

This PR removes the *reasons* those lookups exist, so the Swift side can make ordinary compiler-checked calls. It stands on its own — the encapsulation and the wrong nullability annotation are worth fixing regardless of whether the Swift migration ships.

## What

**1. Hold the JS error controller privately behind a Foundation-typed facade** (`b25eb836`)

`NewRelicAgentInternal.h` declared `@property JSErrorController* jsErrorController` behind a forward `@class`. `JSErrorController` is a *Swift* type, and Clang builds the ObjC half of the framework as its own module with no visibility into the Swift half — so it only ever sees an incomplete type and the Swift importer drops the property entirely. KVC was the only way in.

`SessionReplayManager` already demonstrates the right shape: also a Swift type, but it appears only in the `.m` (which imports the generated `NewRelic-Swift.h`) while the header exposes plain behavioural methods. Same treatment here — the property moves into the `.m`'s existing class extension (keeping its `atomic` accessors, so every `self.jsErrorController` body reference is untouched), and the header gains:

```objc
- (BOOL) recordJavascriptErrorWithName:(NSString*)name message:(NSString*)message
                            stackTrace:(NSString*)stackTrace isFatal:(BOOL)isFatal
                  additionalAttributes:(NSDictionary* _Nullable)additionalAttributes;
```

`+[NewRelic recordJavascriptError:...]` forwards to it and keeps its own shutdown and feature-flag checks, so the public `BOOL` contract is unchanged. `jsErrorController` is now referenced nowhere outside `NewRelicAgentInternal.m`.

**2. Name `+harvestController` for Swift** (`509d63b3`)

Because the method name matches its class name, Swift's importer treats it as a factory method and maps it to `init()` — so `NRMAHarvestController()` from Swift allocates a fresh, never-initialized instance instead of returning the `@synchronized` singleton, silently and with no diagnostic. `NS_SWIFT_NAME(shared())` keeps the singleton reachable. ObjC callers are unaffected, and no Swift code calls it today.

**3. Complete the Swift `NRLOG_*` helper families** (`1c28811e`)

Of the agent-log family only `NRLOG_AGENT_DEBUG` existed, and `NRLOG_*_ATTRS` was absent entirely, so Swift callers hand-wrote the 5-argument `NRLogger.log(...)` call — `NewRelic.swift` alone does this 34 times. Adds `NRLOG_AGENT_ERROR/WARNING/INFO/VERBOSE/AUDIT` plus the six `_ATTRS` variants.

**4. Fix a wrong nullability annotation on `-currentSessionId`** (`8a03e307`)

`currentSessionId` sits inside this header's `NS_ASSUME_NONNULL` region, so it promises non-nil, but it returns `[self agentConfiguration].sessionIdentifier`, which is only assigned in `-onSessionStart` and is therefore nil until a session begins. Swift trusts the annotation, so a nil crosses the bridge as a seemingly valid `String`. Marked `_Nullable`.

The one Swift caller, `NRMASessionReplay.swift:535`, already reads `agent?.currentSessionId() ?? "unknown_session"` — that `??` currently only covers `agent` being nil via optional chaining, so it compiles unchanged and additionally stops treating a nil session id as valid. The 5 ObjC callers are unaffected.

## Testing

- iOS, tvOS and watchOS all build (this PR edits `#if TARGET_OS_IOS`-guarded code in a header, so all three matter)
- Full `Agent_Tests` suite: 800 tests, only the 7 known network-dependent failures (`NRMASessionExclusivityWithDelegateTests` ×5, `NRMAWKNavigationDelegateBaseTest` ×2)
- **New tests are the first coverage of `+[NewRelic recordJavascriptError:...]` at all** — every existing test drives `JSErrorController` directly and so never crosses the public API or this facade, which is how the KVC went unnoticed. They assert the `BOOL` contract in both directions and, via a mock, that arguments route verbatim. **Verified decisive by mutation**: removing the routing makes the test fail with its intended diagnostic; restoring it passes.
- The added logging helpers were **verified mechanically, not by eye** — a parse of both `NRLogger.h` and `NRLogger.swift` confirms every new helper matches its ObjC macro's `(level, agentLogsOn)` pair, and that Swift now has no missing counterpart.
- End-to-end through the test-harness capture server: a JS error recorded via the public API still reaches `/mobile/errors` with a passing payload. (That run also confirms #894 — the same payload previously failed the verifier's `platform == ReactNative` check.)

## Deliberately not in this PR

That mechanical parity check surfaced a **pre-existing** discrepancy: the five bare Swift helpers (`NRLOG_ERROR/WARNING/INFO/VERBOSE/AUDIT`) pass `withAgentLogsOn: true`, while their identically-named ObjC macros pass `false`. Since that flag decides whether a line uploads at its own remote level or only at Debug, the two languages currently ship **different data for the same call**. Correcting it is a behaviour change rather than an addition, so it belongs in its own review — only one call site in the agent uses any of the five today (`NRMASessionDurationManager.swift:77`). Until then, prefer the explicit `NRLOG_AGENT_*` helpers, which mean the same thing before and after that correction.

## Relationship to #837

#837 consumes this: with these landed it replaces its remaining `value(forKey:)` lookups with direct calls and deletes the accompanying workaround comments. Sequenced this way, #837 stays a contained port that *removes* hacks rather than shipping them, and these fixes get reviewed on their own merits.


Agent/General/NewRelicAgentInternal.m
```diff
@@ -101,6 +101,15 @@ @interface NewRelicAgentInternal() {
 @property(nonatomic, strong) NRMAAppInstallMetricGenerator* appInstallMetricGenerator;
 @property(nonatomic, strong) NRMAAppUpgradeMetricGenerator* appUpgradeMetricGenerator;
 
+#if TARGET_OS_IOS
+// Moved here from NewRelicAgentInternal.h. JSErrorController is a Swift type, and this file can
```
cdillard-NewRelic: Can you please remove some of the unneeded comments? I think its best that we manually make sure any comments Claude leaves in are relevant. Its a balance Im still working on as well so no worries.

diegomtz5: Agreed, I removed everything that wasn't needed.

cdillard-NewRelic (approved):

